PR #1, opened 2019-05-29
Add HTTP workload

jmencak:
This is a work-in-progress  "publish early" PR to give a preview of my approach to porting the current [HTTP workload](https://github.com/jmencak/http-ci-tests) to the openshift-scale framework.

Only two external images are currently in use. The workload generator with the mb client and results post-processing `quay.io/jmencak/centos-stress` and the backend nginx pods `quay.io/jmencak/nginx`.  Both are planned to be moved to https://github.com/openshift-scale/images and build from quay.io openshift-scale org.

### DIFF
--- a/workloads/files/workload-http-script-cm.yml
+++ b/workloads/files/workload-http-script-cm.yml
@@ -1,0 +1,1232 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: scale-ci-workload-script
+data:
+  run.sh: |
+    #!/bin/bash
+    set -e
+    # pbench Configuration
+    mkdir -p /var/lib/pbench-agent/tools-default/
+    echo "${USER_NAME:-default}:x:$(id -u):0:${USER_NAME:-default} user:${HOME}:/sbin/nologin" >> /etc/passwd
+    echo "" > /var/lib/pbench-agent/tools-default/oc
+    echo "workload" > /var/lib/pbench-agent/tools-default/label
+    if [[ -v ENABLE_PBENCH_AGENTS ]]; then
+      echo "" > /var/lib/pbench-agent/tools-default/disk
+      echo "" > /var/lib/pbench-agent/tools-default/iostat
+      echo "" > /var/lib/pbench-agent/tools-default/mpstat
+      echo "" > /var/lib/pbench-agent/tools-default/perf
+      echo "" > /var/lib/pbench-agent/tools-default/pidstat
+      master_nodes=`oc get nodes --show-labels -l pbench_agent=true,node-role.kubernetes.io/master= | grep master | awk '{print $1}'`
+      for node in $master_nodes; do
+        echo "master" > /var/lib/pbench-agent/tools-default/remote@$node
+      done
+      infra_nodes=`oc get nodes --show-labels -l pbench_agent=true,node-role.kubernetes.io/infra= | grep infra | awk '{print $1}'`
+      for node in $infra_nodes; do
+        echo "infra" > /var/lib/pbench-agent/tools-default/remote@$node
+      done
+      worker_nodes=`oc get nodes --show-labels -l pbench_agent=true,node-role.kubernetes.io/worker= | grep worker | awk '{print $1}'`
+      for node in $worker_nodes; do
+        echo "worker" > /var/lib/pbench-agent/tools-default/remote@$node
+      done
+    fi
+    source /opt/pbench-agent/profile
+    # End pbench Configuration
+
+    # Test Configuration
+    echo "$(date -u) Configuring HTTP test"
+    test_dir=/tmp/http-ci-tests
+    export HOME=$test_dir
+    mkdir -p $test_dir/content/quickstarts/{nginx,stress}
+    install -m 644 /root/workload/server-*.yaml $test_dir/content/quickstarts/nginx
+    install -m 644 /root/workload/stress-*.yaml $test_dir/content/quickstarts/stress
+    install -m 755 /root/workload/*.sh $test_dir
+    #SERVER_RESULTS=${SERVER_RESULTS:-scp://root@${OCP_NODE_NAME}:/var/lib/pbench-agent}
+    SERVER_RESULTS=${SERVER_RESULTS:-kcp://scale-ci-tooling/${OCP_POD_NAME}:/var/lib/pbench-agent}
+    # End Test Configuration
+
+    # Start of Test Code
+    echo "$(date -u) Running HTTP test"
+    cd $test_dir
+    ./http-test.sh all
+    echo "$(date -u) Completed HTTP test"
+    # End of Test Code
+
+  env.sh: |
+    # Test configuration: a label to append to pbench directories and mb result directories.
+    export TEST_CFG=${TEST_CFG:-1router}
+    # A list of paths to k8s configuration files.  Colon-delimited for Linux.
+    export KUBECONFIG=${KUBECONFIG:-/root/.kube/config}
+    # Use benchmarking and performance analysis framework Pbench.
+    export PBENCH_USE=${PBENCH_USE:-false}
+    # Pbench scraper/wedge usage for r2r analysis.
+    export PBENCH_SCRAPER_USE=${PBENCH_SCRAPER_USE:-false}
+    # Clear results from previous pbench runs.
+    export CLEAR_RESULTS=${CLEAR_RESULTS:-true}
+    # Move the benchmark results to a pbench server.
+    export MOVE_RESULTS=${MOVE_RESULTS:-false}
+    # HTTP-test specifics
+    # Endpoint for collecting results from workload generator node(s).  Keep empty if copying is not required.
+    # - scp://[user@]server:[path]
+    # - kcp://[namespace/]pod:[path]
+    # Note: when PBENCH_USE=true, path is overriden by benchmark_run_dir environment variable
+    export SERVER_RESULTS=${SERVER_RESULTS}
+    # Path to private key when using scp:// to copy results to SERVER_RESULTS.  It is mounted to workload generator container(s).
+    export SERVER_RESULTS_SSH_KEY=${SERVER_RESULTS_SSH_KEY:-/root/.ssh/id_rsa}
+    # How many workload generators to use.
+    export LOAD_GENERATORS=${LOAD_GENERATORS:-1}
+    # Load-generator nodes described by an extended regular expression (use "oc get nodes" node names).
+    # If unset/empty, do not pin the workload generators to any nodes.
+    export LOAD_GENERATOR_NODES=${LOAD_GENERATOR_NODES:-}
+    # Number of projects to create for each type of application (4 types currently).
+    export CL_PROJECTS=${CL_PROJECTS:-10}
+    # Number of templates to create per project.
+    export CL_TEMPLATES=${CL_TEMPLATES:-1}
+    # Run time of individual HTTP test iterations in seconds.
+    export RUN_TIME=${RUN_TIME:-120}
+    # Maximum delay between client requests in ms.
+    export MB_DELAY=${MB_DELAY:-0}
+    # Use TLS session reuse.
+    export MB_TLS_SESSION_REUSE=${MB_TLS_SESSION_REUSE:-true}
+    # HTTP method to use for backend servers (GET/POST).
+    export MB_METHOD=${MB_METHOD:-GET}
+    # Backend server (200 OK) response document length.
+    export MB_RESPONSE_SIZE=${MB_RESPONSE_SIZE:-1024}
+    # Body length of POST requests in characters for backend servers.
+    export MB_REQUEST_BODY_SIZE=${MB_REQUEST_BODY_SIZE:-1024}
+    # Perform the test for the following (comma-separated) route terminations: mix,http,edge,passthrough,reencrypt
+    export ROUTE_TERMINATION=${ROUTE_TERMINATION:-mix}
+    # Run only a single HTTP test to establish functionality.  It also overrides CL_PROJECTS and CL_TEMPLATES.
+    export SMOKE_TEST=${SMOKE_TEST:-false}
+    # Delete all namespaces with application pods, services and routes created for the purposes of HTTP tests.
+    export NAMESPACE_CLEANUP=${NAMESPACE_CLEANUP:-true}
+    # HTTP workload generator container image
+    export HTTP_STRESS_CONTAINER_IMAGE=${HTTP_STRESS_CONTAINER_IMAGE:-quay.io/openshift-scale/http-stress}
+    # HTTP server container image
+    export HTTP_SERVER_CONTAINER_IMAGE=${HTTP_SERVER_CONTAINER_IMAGE:-quay.io/openshift-scale/nginx}
+
+  http-test.sh: |
+    #!/bin/bash
+
+    ProgramName=${0##*/}
+
+    . ./env.sh
+
+    ### Global variables ###########################################################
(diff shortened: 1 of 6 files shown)


jmencak:
@akrzos, still very much a WiP, but a high-level feedback if I'm on the right track welcome, thx.

workloads/http.yml
```diff
@@ -0,0 +1,121 @@
...
+  gather_facts: true
+  remote_user: "{{orchestration_user}}"
+  vars_files:
+    - vars/http.yml
+  vars:
+    workload_job: "http"
+    enable_pbench_agents: true
+    workload_job_privileged: true
+    workload_job_node_selector: true
+    workload_job_taint: true
```
akrzos: I think we can remove the following vars here since they should be defined under the vars file `workloads/vars/http.yml` although it does look like one is missing (workload_job_node_selector) in the vars file.

```
    enable_pbench_agents: true
    workload_job_privileged: true
    workload_job_node_selector: true
    workload_job_taint: true
```

workloads/vars/http.yml
```diff
@@ -0,0 +1,37 @@
...
+ansible_public_key_file: "{{ lookup('env', 'PUBLIC_KEY')|default('~/.ssh/id_rsa.pub', true) }}"
+ansible_private_key_file: "{{ lookup('env', 'PRIVATE_KEY')|default('~/.ssh/id_rsa', true) }}"
+
+orchestration_user: "{{ lookup('env', 'ORCHESTRATION_USER')|default('root', true) }}"
+###############################################################################
+# RHCOS http workload variables.
+###############################################################################
+
+# Container image in use
+workload_image: "{{ lookup('env', 'WORKLOAD_IMAGE')|default('quay.io/openshift-scale/scale-ci-workload', true) }}"
```
akrzos, line 14: I believe we are missing the option for the var to use a nodeselector for the workload job here

```
# Use nodeselector to place workload job on specific node
workload_job_node_selector: "{{ lookup('env', 'WORKLOAD_JOB_NODE_SELECTOR')|default('', true)|bool }}"
```

workloads/files/workload-http-script-cm.yml
```diff
@@ -0,0 +1,1218 @@
...
+    # Clear results from previous pbench runs.
+    export CLEAR_RESULTS=${CLEAR_RESULTS:-true}
+    # Move the benchmark results to a pbench server.
+    export MOVE_RESULTS=${MOVE_RESULTS:-false}
+    # HTTP-test specifics
+    # Endpoint for collecting results from workload generator node(s).  Keep empty if copying is not required.
+    # - scp://[user@]server:[path]
+    # - kcp://[namespace/]pod:[path]
+    # Note: when PBENCH_USE=true, path is overriden by benchmark_run_dir environment variable
+    export SERVER_RESULTS=${SERVER_RESULTS}
```
akrzos, line 73: I think we need to set the port here but I'll continue to play with this tomorrow to determine what I am doing wrong.

akrzos (commented):
I'll need more time to run this however I think I found a missing env var for the workload node selector.

I have ran the workload however the centos-stress pod ran into this issue:

```
+ ssh -o StrictHostKeyChecking=false -o UserKnownHostsFile=/dev/null -o ConnectTimeout=20 -i /opt/stress/.ssh/id_rsa -T root@ip-10-0-11-148.us-west-2.compute.internal
Warning: Permanently added 'ip-10-0-11-148.us-west-2.compute.internal,10.0.11.148' (ECDSA) to the list of known hosts.
Permission denied, please try again.
Permission denied, please try again.
Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).
+ scp -o StrictHostKeyChecking=false -o UserKnownHostsFile=/dev/null -o ConnectTimeout=20 -i /opt/stress/.ssh/id_rsa -rp mb-centos-stress-hfdqg-1 root@ip-10-0-11-148.us-west-2.compute.internal:/var/lib/pbench-agent/pbench-user-benchmark_0004r-001cpt-0000d_ms-001ka-ytlsru-120s-mix-1router_2019.05.31T01.55.31
Warning: Permanently added 'ip-10-0-11-148.us-west-2.compute.internal,10.0.11.148' (ECDSA) to the list of known hosts.
Permission denied, please try again.
Permission denied, please try again.
Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).
lost connection
+ test 1 -ne 0
+ warn 'couldn'\''t scp test results, sleep inf, please debug'
+ fail 'run: couldn'\''t scp test results, sleep inf, please debug'
+ echo run: 'couldn'\''t' scp test results, sleep inf, please debug
run: couldn't scp test results, sleep inf, please debug
```

I believe I have a difference in my environment than the environment your running in which I haven't populated the root user's authorized_keys.  I'll investigate more tomorrow.  Thanks for the huge effort so far Jiri!

jmencak:
Thanks for all the comments and your time, Alex! The environment cleanup is definitely on my TODO and by no means considered ready for public consumption, hence the WiP.  The environment is very likely why you're seeing the problems.  What I was more interested in hearing is whether the overall structure of the PR -- the integration with the existing code -- looks OK to you to prevent any waste of time by further improving and cleaning it up.

workloads/vars/http.yml
jmencak: Does this need to be configurable for every workload?  Also, the variable name seems to be slightly misleading.  `workload_job_node_selector` is a boolean (not an actual node selector) that makes the workload job to be placed on nodes labelled `node-role.kubernetes.io/pbench`.

workloads/http.yml
jmencak: Good catch! The env cleanup, is on my TODO.  As I said below, do we need to have these for every workload in `workloads/vars/<workload>.yml`?

workloads/files/workload-http-script-cm.yml
jmencak: I believe the issue above might have been caused by the container running on a worker node, rather than the pbench node and not having the ssh keys configured properly for the results to be copied back.  In any case, perhaps using kcp:// by default and `OCP_POD_NAME` will be a safer option for this framework.  Need to think about this.

workloads/http.yml
akrzos: Well I think it would be convenient if all environmentally defined variables are provided in the vars file `workloads/vars/http.yml` with the exception of non-environmentally defined (or vars that we don't need provided from jenkins) could be defined at the top of the workload playbooks.  A good example of this is the test job which is running a sshd daemon in the workload job container, thus would always require the node selector to be placed on our workload/pbench node which has a public ip and can thus be reachable via sshd.

In regards to if every workload would require those vars (enable node selector/taint and privileged) to make the goal of supporting environments not provisioned by our install automation we will need to provide options to run without those items switched on.

workloads/vars/http.yml
akrzos: Yeah it is not the best name for a variable, mostly because I wanted to be able to achieve what you mentioned however at the time I didn't invest further into making the node selector completely configurable.  At the time a boolean enabling/disabling the node selector for our pbench/workload node was easiest.  If you'd like to add that functionality, lets get another pr.

workloads/http.yml
jmencak: I've dropped most of these from the environment in `workloads/vars/http.yml` now, only `ENABLE_PBENCH_AGENTS` is left and removed `enable_pbench_agents` from `workloads/workloads/http.yml`.  `workload_job_privileged` is false now, so, we're only left with `workload_job_node_selector` and `workload_job_taint` set to true.  `workload_job_node_selector` can be confusing, as it is a `boolean` which forces a node with a  pbench label to be used.  Also not very happy about the `workload_job_taint` naming.  I will look into ways of making these false so that the variables don't need to be exposed at all.

akrzos: I will be playing with this pr more later today for more feedback.  My initial thoughts for those vars was that turning them off would be a tradeoff between workload isolation and ability to run the workloads on environments we didn't provision.  I hope that helps explain otherwise we can pair up over bluejeans sometime to discuss.  Also I am up for better names of course!

jmencak: Could be just my lack of understanding what's the purpose for having 2 variables.  From what I can see 1 variable could easily replace them.  Something like `run_on_pbench_node` but that's not the variable name I'm suggesting.  Variable naming is one of the the hardest problems in computer science ;) Let's discuss offline.

akrzos:
@chaitanyaenr and @sjug PTAL and review.  Thanks!

workloads/vars/http.yml
```diff
@@ -0,0 +1,33 @@
...
+pbench_ssh_public_key_file: "{{ lookup('env', 'PBENCH_SSH_PUBLIC_KEY_FILE')|default('~/.ssh/id_rsa.pub', true) }}"
+enable_pbench_agents: "{{ lookup('env', 'ENABLE_PBENCH_AGENTS')|default(false, true)|bool }}"
+pbench_server: "{{ lookup('env', 'PBENCH_SERVER')|default('', true) }}"
+
+# Other variables for workload tests
+scale_ci_results_token: "{{ lookup('env', 'SCALE_CI_RESULTS_TOKEN')|default('', true) }}"
+job_completion_poll_attempts: "{{ lookup('env', 'JOB_COMPLETION_POLL_ATTEMPTS')|default(1000, true)|int }}"
+
+# HTTP workload specific parameters
+http_env_vars: [ "TEST_CFG", "PBENCH_USE", "PBENCH_SCRAPER_USE", "CLEAR_RESULTS", "MOVE_RESULTS", "SERVER_RESULTS",
```
sjug, line 30: Are these set only in environment variables? Are there some defaults?

jmencak: Sure, the defaults come bundled with `workloads/files/workload-http-script-cm.yml`, `env.sh`.  This is intentional to keep this as much as "drop-in" with the "upstream" [http-ci-tests](https://github.com/jmencak/http-ci-tests) as possible.

workloads/files/workload-http-script-cm.yml
```diff
@@ -0,0 +1,1217 @@
...
+      done
+    fi
+    source /opt/pbench-agent/profile
+    # End pbench Configuration
+
+    # Test Configuration
+    echo "$(date -u) Configuring HTTP test"
+    test_dir=/tmp/http-ci-tests
+    mkdir -p $test_dir/content/quickstarts/{nginx,stress}
+    install -m 644 /root/workload/server-*.yaml $test_dir/content/quickstarts/nginx
```
akrzos, line 41: The latest workload container image has a newer cluster loader binary which no longer requires us to juggle files around as we can use absolute and relative paths to ensure the correct configuration and test files are used now.

workloads/files/workload-http-script-cm.yml
```diff
@@ -0,0 +1,1217 @@
...
+
+    # Start of Test Code
+    echo "$(date -u) Running HTTP test"
+    cd $test_dir
+    ./http-test.sh all
+    echo "$(date -u) Completed HTTP test"
+    # End of Test Code
+
+  ###################################################################################################
+  env.sh: |
```
akrzos, line 55: I looked in a bit more on how vars are moving here, currently AIUI we have:

User/Jenkins provides values via env vars -> Ansible reads env vars -> creates env configmap -> configmap is read by another env.sh script ( and establishes defaults) -> used in the script itself

I understand this design is to maintain "drop-in" compatibility with the upstream http tests, however is it at all possible to establish defaults at when Ansible reads the env vars thus eliminating the need for the env.sh script?  I am ok with it if there isn't in order to make "drop-in" with upstream easier and faster, so this is just an idea not a necessity.

workloads/files/workload-http-script-cm.yml
```diff
@@ -0,0 +1,1217 @@
...
+          - name: IDENTIFIER
+            value: ${IDENTIFIER}
+          - name: NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
+          envFrom:
+          - configMapRef:
+              name: wlg-env
+          image: quay.io/jmencak/centos-stress
```
akrzos: I think we should consider making this image configurable in the event we want to more quickly use a different stress container image.

workloads/files/workload-http-script-cm.yml
```diff
@@ -0,0 +1,1217 @@
...
+            optional: true
+          name: targets
+        - name: ssh-key
+          secret:
+            items:
+            - key: wlg-ssh-key
+              path: id_rsa
+            optional: true
+            secretName: wlg-ssh-key
+            # yaml/json doesn't support octal: 400 = (4 * 8^2) + (4 * 8^1) + (4 * 8^0) => 256
```
akrzos: I am not sure I agree with this comment - https://github.com/openshift-scale/workloads/blob/master/workloads/templates/workload-job.yml.j2#L86

however perhaps the context in how the reference above is rendered (j2 -> yaml), might be providing the value for defaultMode in the correct "type"

workloads/files/workload-http-script-cm.yml
```diff
@@ -0,0 +1,1217 @@
...
+        selector:
+          name: nginx-http-${IDENTIFIER}
+        template:
+          metadata:
+            labels:
+              name: nginx-http-${IDENTIFIER}
+              test: http
+          spec:
+            containers:
+            - image: quay.io/jmencak/nginx
```
akrzos: Like wise to above comment regarding container image I think we should consider allowing this and the other templates with nginx images to be configured.

workloads/files/workload-http-script-cm.yml
jmencak: The point of copying is to have a drop-in replacement for the "upstream" [http-ci-tests](https://github.com/jmencak/http-ci-tests).  Mounting a read-only ConfigMap could be an option, but this would complicate things.

workloads/files/workload-http-script-cm.yml
jmencak: Agreed.

workloads/files/workload-http-script-cm.yml
jmencak: Agreed.

workloads/files/workload-http-script-cm.yml
jmencak: This really comes down to where you want the interface to be.  I've actually had the env variables defaulted by Ansible, but that was adding a lot of unnecessary code and not providing sane defaults at the level of the scripts themselves.  Also, as you mention, the scripts would no longer be self-contained unless you kept env.sh, which would lead to code duplication.  Neither approach is "great" in my view, but my approach at least preserves the "drop-in" functionality.  Will think about this more, but not sure I'll figure out a better solution.

workloads/files/workload-http-script-cm.yml
akrzos: Got it, I forgot how one requirement here is to allow modification of the files.  Essentially the http workload performs the templateing of the files thus a configmap really is just for "shipping" the files into the container image.

workloads/files/workload-http-script-cm.yml
jmencak: Interesting, didn't work for me when I tested, but will double check and change to `0400` if that really works.

workloads/files/workload-http-script-cm.yml
akrzos: For current purposes, "drop-in" functionality is more important (at least in my opinion) so we can continue to think up ideas here and if anything seems to fit the bill for both of those, then we can propose a pr later for the change.

akrzos (commented):
Everything looks good, I'm going to run the smoke test to confirm it is running in a few moments.

PR #5, opened 2019-05-29
Wait until the pbench agent daemonsets are ready

akrzos:
Although the wait time is usually negligible there is the potential to start a
test before the pbench agent daemonsets are up and running.  This p/r adds a
task to wait until the daemonsets are ready.  Additionally, there is a task
added to wait for scale-ci-tooling task deletion to complete.


sjug:
Looks good, same comment as https://github.com/openshift-scale/workloads/pull/6 :+1:

akrzos:
Thanks, will one of you merge it?

chaitanyaenr (approved):
/lgtm

chaitanyaenr:
@akrzos can we increase the delay to something more than 1 sec like @sjug mentioned?

akrzos:
Sure what delay and retry combo do you all suggest?

jmencak:
Do we need to increase the delay?  Isn't the call blocking anyway?  Is 1s too much?

akrzos:
The agents come up pretty quick for me at least within 1-2s, but maybe I should try and clear the images first and see how long it typically takes for them to be pulled and started.

chaitanyaenr:
I think the agents will come up quickly like @akrzos mentioned. It should be the case even when we don't the image pulled down to the node as the image size has been drastically reduced. Think 1 sec polling should be fine.

akrzos:
Ok with a cluster without the `scale-ci-workload` image took 10 attempts before the playbook found the daemonset up and running.  On a "warm" cluster it took 2 attempts.  The task `Wait until scale-ci-tooling project is deleted` took the longest time on the warm cluster so I figured best here to just increase the delay from 1s to 2s so we have a 2 minute window for both of those waiting tasks.  Do note that the 2 minute window is the minimum amount of time and connection overhead will lengthen this a tiny bit (unless you have a connection with greater than "normal" latency)

chaitanyaenr:
2s delay sounds good, merging the PR.

PR #6, opened 2019-05-30
Add ability to wait for a workload to complete

akrzos:


sjug:
Looks good other than the aggressive retry delay. If we're waiting a minute I'd rather something like 20 retries every 5s.

jmencak:
Broke my rule of /lgtm without testing again.
/lgtm cancel
```
[DEPRECATION WARNING]: evaluating wait_for_job_complete as a bare variable, this behaviour will go away and you might need to add |bool to the expression in the future. Also see 
CONDITIONAL_BARE_VARS configuration toggle.. This feature will be removed in version 2.12. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
 [WARNING]: conditional statements should not include jinja2 templating delimiters such as {{ }} or {% %}. Found: {{ lookup('env', 'WAIT_FOR_JOB_COMPLETE')|default(true, true)|bool }}
```
and more importantly, don't you want to check for job *completion*, rather than job *configuration* to finish?  If so, wouldn't checking a job status be a safer way of checking the status rather than trawling through job logs?

akrzos:
> wait_for_job_complete



> Broke my rule of /lgtm without testing again.
> /lgtm cancel
> 
> ```
> [DEPRECATION WARNING]: evaluating wait_for_job_complete as a bare variable, this behaviour will go away and you might need to add |bool to the expression in the future. Also see 
> CONDITIONAL_BARE_VARS configuration toggle.. This feature will be removed in version 2.12. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
>  [WARNING]: conditional statements should not include jinja2 templating delimiters such as {{ }} or {% %}. Found: {{ lookup('env', 'WAIT_FOR_JOB_COMPLETE')|default(true, true)|bool }}
> ```

hmm let me look into this warnings, what version of ansible are you running?

> and more importantly, don't you want to check for job _completion_, rather than job _configuration_ to finish? If so, wouldn't checking a job status be a safer way of checking the status rather than trawling through job logs?

Yes, let me see if this can be satisfied with just checking job status instead of `oc logs -f` and looking for an identifier in the log output.  I had originally chose the first identifier we typically expect the script to output that should be available for all jobs and the `-f` would essentially cause the command to block until completion however I think this will fail on a very long running job or a job that the oc logs command drops its connection and that identifier was actually already in the log output too.

For job status this might be tricky since depending on the actual state of the job, depends on what is include in the json output from oc command:
```
root@ip-172-31-32-21: ~/openshift-scale/workloads # oc get job scale-ci-nodevertical -n scale-ci-tooling -o json | jq '.status'
{
  "active": 1,
  "startTime": "2019-05-30T11:25:40Z"
}
root@ip-172-31-32-21: ~/openshift-scale/workloads # oc get job scale-ci-nodevertical -n scale-ci-tooling -o json | jq '.status'
{
  "completionTime": "2019-05-30T11:29:07Z",
  "conditions": [
    {
      "lastProbeTime": "2019-05-30T11:29:07Z",
      "lastTransitionTime": "2019-05-30T11:29:07Z",
      "status": "True",
      "type": "Complete"
    }
  ],
  "startTime": "2019-05-30T11:25:40Z",
  "succeeded": 1
}
```

jmencak:
> hmm let me look into this warnings, what version of ansible are you running?

ansible 2.8.0

as for the completion, I had something like this in mind:
```
until: "{{ job_scale_ci_http.stdout | from_json | json_query('status.succeeded') == 1 }}"
```

akrzos:
> > hmm let me look into this warnings, what version of ansible are you running?
> 
> ansible 2.8.0
> 

Got it I think I have the warning addressed now.

> as for the completion, I had something like this in mind:
> 
> ```
> until: "{{ job_scale_ci_http.stdout | from_json | json_query('status.succeeded') == 1 }}"
> ```

Yup I think this will work out better than just repeatedly attempting to tail the logs until completion.  I extended this for if the job fails in the p/r.  Thanks again!

workloads/mastervertical.yml
```diff
@@ -63,15 +63,15 @@
...
-      block:
-        - name: Ensure any stale scale-ci-workload job is deleted
-          shell: |
-            oc delete --wait job scale-ci-mastervertical -n scale-ci-tooling
-          ignore_errors: true
-      when:
-        - scale_ci_tooling_ns_exists.rc == 0
+    - name: Ensure any stale scale-ci-mastervertical job is deleted
+      shell: |
+         oc delete job scale-ci-mastervertical -n scale-ci-tooling
```
jmencak: Indentation, 3 vs 2 spaces.

workloads/mastervertical.yml
```diff
@@ -108,10 +106,20 @@
...
-    # Might need retires to account for api server cert rotation / connection breaking
-    # Wait for job start, then wait for complete
-    # oc get job scale-ci-mastervertical -n scale-ci-tooling -o json | jq '.status.active'
-    # oc get job scale-ci-mastervertical -n scale-ci-tooling -o json | jq '.status.succeeded'
-    # - name: Wait for job to complete
-    #   shell: |
-    #     oc logs -f job/scale-ci-mastervertical -n scale-ci-tooling
+    - name: Poll (2m) until job is active
+      shell: |
+         oc get job scale-ci-mastervertical -n scale-ci-tooling -o json
```
jmencak: Indentation.

workloads/mastervertical.yml
```diff
@@ -108,10 +106,20 @@
...
+      register: job_json
+      retries: 60
+      delay: 2
+      until: job_json.stdout | from_json | json_query('status.active==`1`')
+
+    - name: Poll until job complete
+      shell: |
+         oc get job scale-ci-mastervertical -n scale-ci-tooling -o json
```
jmencak: Indentation.

workloads/mastervertical.yml
```diff
@@ -108,10 +106,20 @@
...
         oc replace --force -n scale-ci-tooling -f "{{ansible_user_dir}}/scale-ci-tooling/workload-job.yml"
 
-    # Might need retires to account for api server cert rotation / connection breaking
-    # Wait for job start, then wait for complete
-    # oc get job scale-ci-mastervertical -n scale-ci-tooling -o json | jq '.status.active'
-    # oc get job scale-ci-mastervertical -n scale-ci-tooling -o json | jq '.status.succeeded'
-    # - name: Wait for job to complete
-    #   shell: |
-    #     oc logs -f job/scale-ci-mastervertical -n scale-ci-tooling
+    - name: Poll (2m) until job is active
```
jmencak: I suggest dropping the (2m).  60*2 is 120s, but the `oc` command usually takes a couple of seconds to return a result itself, therefore in reality this can be a lot more than 2m.

workloads/mastervertical.yml
```diff
@@ -108,10 +106,20 @@
...
+      shell: |
+         oc get job scale-ci-mastervertical -n scale-ci-tooling -o json
+      register: job_json
+      retries: 60
+      delay: 2
+      until: job_json.stdout | from_json | json_query('status.active==`1`')
+
+    - name: Poll until job complete
```
jmencak: Until job *is* complete?  Consistency with the previous code.

jmencak (commented):
/lgtm
only nits. Much better than the previous version, stealing the code for the HTTP workload, thanks!

akrzos:
Adjusted for the last set of feedback.  I think it is gtg now but please review one more time.

workloads/test.yml
```diff
@@ -67,15 +67,15 @@
...
-      block:
-        - name: Ensure any stale scale-ci-workload job is deleted
-          shell: |
-            oc delete job scale-ci-workload -n scale-ci-tooling
-          ignore_errors: true
-      when:
-        - scale_ci_tooling_ns_exists.rc == 0
+    - name: Ensure any stale scale-ci-test job is deleted
+      shell: |
+         oc delete job scale-ci-test -n scale-ci-tooling
```
jmencak: 3 spaces.  I didn't point out all of these in the PR, sorry.

akrzos: Ah, no worries I think I got all of them.  Maybe we can get a linter bot/job to find these more easily.

jmencak: Yep, some ci with a linter would be nice.

chaitanyaenr (approved):
LGTM

akrzos:
Can we have this merged now?

PR #13, opened 2019-06-04
No longer need to move files around for Cluster Loader

akrzos:
We no longer need to move files to a directory structure specific for cluster
loader.  Also the files referenced by Cluster Loader can now be yaml instead
of json.


workloads/templates/workload-nodevertical-script-cm.yml.j2
```diff
@@ -61,7 +53,7 @@ data:
             - num: {{total_pod_count.stdout|int}}
               image: {{nodevertical_pod_image}}
               basename: pausepods
-              file: pod-pause-labeled-nodes.json
+              file: pod-pause-labeled-nodes.yaml
```
sjug, line 56: This file will be in the same `/root/workload/` directory as the CL config?

akrzos: Yup it is placed in the same directory as the cluster loader config file.

akrzos: I changed it so all three workloads consistently refer to their associated files in the same manner (`file: file_name.yaml` instead of mixed with `file: ./file_name.yaml`)

akrzos:
@chaitanyaenr let me know if you have any feedback, the latest container images has the changes to support this at this time.

chaitanyaenr (approved):
@akrzos LGTM.

Issue #16, opened 2019-06-06
Separate readme files for each workload


chaitanyaenr:
We need to have separate readme files for each workload supported as the length of the readme might be too big as we add more workloads to the repo like @akrzos suggested.

akrzos:
@chaitanyaenr @ekuric @jmencak @Deepthidharwar @sjug Any suggestions of documentation "tools" to use here, should we continue to use markdown files in a docs directory, maybe use the wiki feature of github and/or use some sort of other docs tool?

My opinion is that the markdown files will likely be the lowest impact to current work efforts and we could just create an organized structure in a docs directory to describe each workload and its vars.

jmencak:
Totally agree with the issue.  Markdown files sound good.

ekuric:
+1 , also I would recommend workloads/workloads/storage as separate location for storage tests

akrzos:
At this point we now how a separate docs directory and separate README files per workload.  Closing this issue.

PR #17, opened 2019-06-10
Network tests

akrzos:
4th pass at refactoring of the network tests.
* Pod 2 Pod is working
* Pod 2 Pod with HostNetwork is working with privileged pods
* Svc 2 Svc is working with privileged pods

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,3 +337,132 @@ NAMESPACE_CLEANUP
 HTTP_STRESS_CONTAINER_IMAGE
 HTTP_SERVER_CONTAINER_IMAGE
 ```
+
+## RHCOS Network tests Workload
+
+The RHCOS network tests workload playbook is `workloads/network.yml` and is used to test latency and throughput of the cluster network.
+
+Requirements:
+* All Network tests - labeling nodes
+* HostNetwork - privileged pods
+* HostNetwork - Cloud firewall configuration to permit traffic for ports between 20010-20100 and NETWORK_TEST_UPERF_SSHD_PORT
+
+Running from CLI:
+
+```sh
+$ cp workloads/inventory.example inventory
+$ # Add orchestration host to inventory
+$ # Edit vars in workloads/vars/network.yml or define Environment vars (See below)
+$ time ansible-playbook -vv -i inventory workloads/network.yml
+```
+
+### Environment variables for RHCOS network workload playbook
+
+```
+###############################################################################
+# Ansible SSH variables.
+###############################################################################
+PUBLIC_KEY
+PRIVATE_KEY
+
+ORCHESTRATION_USER
+###############################################################################
+# RHCOS workload variables.
+###############################################################################
+WORKLOAD_IMAGE
+
+WORKLOAD_JOB_NODE_SELECTOR
+WORKLOAD_JOB_TAINT
+WORKLOAD_JOB_PRIVILEGED
+
+KUBECONFIG_FILE
+
+PBENCH_SSH_PRIVATE_KEY_FILE
+PBENCH_SSH_PUBLIC_KEY_FILE
+ENABLE_PBENCH_AGENTS
+PBENCH_SERVER
+
+SCALE_CI_RESULTS_TOKEN
+JOB_COMPLETION_POLL_ATTEMPTS
+
+# Network benchmarks specific parameters:
+NETWORK_TEST_UPERF_IMAGE
+NETWORK_TEST_UPERF_SSHD_PORT
+NETWORK_TEST_UPERF_PORT_RANGE
+NETWORK_TEST_PREFIX
+NETWORK_TEST_PAIRS
+NETWORK_TEST_BASENAME
+NETWORK_TEST_HOSTNETWORK
+NETWORK_TEST_SERVICE
+NETWORK_TEST_TYPE
+NETWORK_TEST_RUNTIME
+NETWORK_TEST_MESSAGE_SIZES
+NETWORK_TEST_PROTOCOL
+NETWORK_TEST_INSTANCES
+NETWORK_TEST_SAMPLES
+NETWORK_TEST_CLEANUP
+```
+
+### Notes for Network testing execution
+
+#### Suggested Pod2Pod Test Configuration
+
+```
+NETWORK_TEST_UPERF_IMAGE="quay.io/openshift-scale/scale-ci-uperf"
+NETWORK_TEST_UPERF_SSHD_PORT=20000
+NETWORK_TEST_UPERF_PORT_RANGE="20010 20109"
+NETWORK_TEST_PREFIX=network
+NETWORK_TEST_PAIRS=1,2,4
+NETWORK_TEST_BASENAME=network
+NETWORK_TEST_HOSTNETWORK=false
+NETWORK_TEST_SERVICE=false
+NETWORK_TEST_TYPE=stream,rr
+NETWORK_TEST_RUNTIME=60
+NETWORK_TEST_MESSAGE_SIZES=64,1024,16384
+NETWORK_TEST_PROTOCOL=tcp,udp
+NETWORK_TEST_INSTANCES=1,2
+NETWORK_TEST_SAMPLES=3
+NETWORK_TEST_CLEANUP=true
+```
+
+#### Suggested Pod2Pod HostNetwork Test Configuration
+
+```
+NETWORK_TEST_UPERF_IMAGE="quay.io/openshift-scale/scale-ci-uperf"
+NETWORK_TEST_UPERF_SSHD_PORT=20000
+NETWORK_TEST_UPERF_PORT_RANGE=""
+NETWORK_TEST_PREFIX=network
+NETWORK_TEST_PAIRS=1
+NETWORK_TEST_BASENAME=network
+NETWORK_TEST_HOSTNETWORK=true
+NETWORK_TEST_SERVICE=false
+NETWORK_TEST_TYPE=stream,rr
+NETWORK_TEST_RUNTIME=60
+NETWORK_TEST_MESSAGE_SIZES=64,1024,16384
+NETWORK_TEST_PROTOCOL=tcp,udp
+NETWORK_TEST_INSTANCES=1,2,4
+NETWORK_TEST_SAMPLES=3
+NETWORK_TEST_CLEANUP=true
+```
+
+Make sure NETWORK_TEST_UPERF_SSHD_PORT is enabled in your cloud environment to permit traffic when using HostNetwork.  HostNetwork testing requires privileged pods.
+
+#### Suggested Service 2 Service Test Configuration
+
+```
+NETWORK_TEST_UPERF_IMAGE="quay.io/openshift-scale/scale-ci-uperf"
+NETWORK_TEST_UPERF_SSHD_PORT=20000
+NETWORK_TEST_UPERF_PORT_RANGE="20010 20109"
+NETWORK_TEST_PREFIX=network
+NETWORK_TEST_PAIRS=1,2,4
+NETWORK_TEST_BASENAME=network
+NETWORK_TEST_HOSTNETWORK=false
+NETWORK_TEST_SERVICE=true
+NETWORK_TEST_TYPE=stream,rr
+NETWORK_TEST_RUNTIME=60
+NETWORK_TEST_MESSAGE_SIZES=64,1024,16384
+NETWORK_TEST_PROTOCOL=tcp,udp
+NETWORK_TEST_INSTANCES=1,2
+NETWORK_TEST_SAMPLES=3
+NETWORK_TEST_CLEANUP=true
+```
--- a/workloads/files/workload-network-script-cm.yml
+++ b/workloads/files/workload-network-script-cm.yml
@@ -1,0 +1,491 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: scale-ci-workload-script
+data:
+  run.sh: |
+    #!/bin/sh
+    set -eo pipefail
+    source /root/workload/functions.sh
+    # pbench Configuration
+    log "Configuring pbench for Network benchmarks"
+    mkdir -p /var/lib/pbench-agent/tools-default/
+    echo "${USER_NAME:-default}:x:$(id -u):0:${USER_NAME:-default} user:${HOME}:/sbin/nologin" >> /etc/passwd
+    echo "" > /var/lib/pbench-agent/tools-default/oc
+    echo "workload" > /var/lib/pbench-agent/tools-default/label
+    if [[ -v ENABLE_PBENCH_AGENTS ]]; then
+      echo "" > /var/lib/pbench-agent/tools-default/disk
+      echo "" > /var/lib/pbench-agent/tools-default/iostat
+      echo "" > /var/lib/pbench-agent/tools-default/mpstat
+      echo "" > /var/lib/pbench-agent/tools-default/perf
+      echo "" > /var/lib/pbench-agent/tools-default/pidstat
+      master_nodes=`oc get nodes -l pbench_agent=true,node-role.kubernetes.io/master= --no-headers | awk '{print $1}'`
+      for node in $master_nodes; do
+        echo "master" > /var/lib/pbench-agent/tools-default/remote@$node
+      done
+      infra_nodes=`oc get nodes -l pbench_agent=true,node-role.kubernetes.io/infra= --no-headers | awk '{print $1}'`
+      for node in $infra_nodes; do
+        echo "infra" > /var/lib/pbench-agent/tools-default/remote@$node
+      done
+      worker_nodes=`oc get nodes -l pbench_agent=true,node-role.kubernetes.io/worker= --no-headers | awk '{print $1}'`
+      for node in $worker_nodes; do
+        echo "worker" > /var/lib/pbench-agent/tools-default/remote@$node
+      done
+    fi
+    source /opt/pbench-agent/profile
+    log "Done configuring pbench Network benchmarks"
+    # End pbench Configuration
+
+    # Start of Test Code
+    log "Running Network benchmarks"
+
+    config_prefix="P2P"
+    cp /root/workload/pod-rc-template.yaml /tmp/pod-rc-template.yaml
+    if [[ "${NETWORK_TEST_HOSTNETWORK}" == "true" ]]; then
+      config_prefix="P2P_HN"
+      envsubst < /root/workload/network-scc.yaml.template > /tmp/network-scc.yaml
+      oc replace --force -f /tmp/network-scc.yaml
+      cp /root/workload/pod-hn-rc-template.yaml /tmp/pod-rc-template.yaml
+    fi
+    # Loop over pair count
+    for pair_count in ${NETWORK_TEST_PAIRS//,/ }
+    do
+      log "Running Network benchmarks with ${pair_count} pair(s)"
+      if [ "${NETWORK_TEST_SERVICE}" == "true" ]; then
+        config_prefix="S2S"
+        PAIR_COUNT=${pair_count} envsubst < /root/workload/s2s-network.yaml.template > /tmp/network.yaml
+        cp /root/workload/svc-template.yaml /tmp/svc-template.yaml
+        # Append required ports to service template
+        for i in `seq ${NETWORK_TEST_UPERF_PORT_RANGE}`
+        do
+          echo -e "    - name: tcp-${i}\n      port: ${i}\n      protocol: TCP" >> /tmp/svc-template.yaml
+          echo -e "    - name: udp-${i}\n      port: ${i}\n      protocol: UDP" >> /tmp/svc-template.yaml
+        done
+      else
+        PAIR_COUNT=${pair_count} envsubst < /root/workload/p2p-network.yaml.template > /tmp/network.yaml
+      fi
+
+      VIPERCONFIG=/tmp/network.yaml openshift-tests run-test "[Feature:Performance][Serial][Slow] Load cluster should load the cluster [Suite:openshift]"
+
+      # Check if clients and servers are ready
+      for rc_num in `seq 0 $((${pair_count} - 1))`
+      do
+        is_replica_ready "uperf-client-${rc_num}"
+        is_replica_ready "uperf-server-${rc_num}"
+      done
+
+      if [ "${NETWORK_TEST_SERVICE}" == "true" ]; then
+        clients=`oc get svc --output=json -n ${NETWORK_TEST_BASENAME}0 --selector="name=client" | jq -r '[.items[].spec.clusterIP] | join(",")'`
+        servers=`oc get svc --output=json -n ${NETWORK_TEST_BASENAME}0 --selector="name=server" | jq -r '[.items[].spec.clusterIP] | join(",")'`
+      else
+        clients=`oc get pods --output=json -n ${NETWORK_TEST_BASENAME}0 --selector="name=client" | jq -r '[.items[].status.podIP] | join(",")'`
+        servers=`oc get pods --output=json -n ${NETWORK_TEST_BASENAME}0 --selector="name=server" | jq -r '[.items[].status.podIP] | join(",")'`
+      fi
+
+      mv ~/.ssh/config ~/.ssh/config.bak || true
+      for host in $(echo "${clients} ${servers}" | sed "s/,/ /g")
+      do
+        HOST=${host} envsubst < /root/workload/ssh_config.template >> ~/.ssh/config
+      done
+      cat ~/.ssh/config.bak >> ~/.ssh/config || true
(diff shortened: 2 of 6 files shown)


akrzos:
DNM (Do Not Merge) yet as I am still working on the Svc2Svc implementation, however please provide some early feedback.  @jmencak  @sjug  @chaitanyaenr

workloads/files/workload-network-script-cm.yml
```diff
@@ -0,0 +1,961 @@
...
+    echo "${USER_NAME:-default}:x:$(id -u):0:${USER_NAME:-default} user:${HOME}:/sbin/nologin" >> /etc/passwd
+    echo "" > /var/lib/pbench-agent/tools-default/oc
+    echo "workload" > /var/lib/pbench-agent/tools-default/label
+    if [[ -v ENABLE_PBENCH_AGENTS ]]; then
+      echo "" > /var/lib/pbench-agent/tools-default/disk
+      echo "" > /var/lib/pbench-agent/tools-default/iostat
+      echo "" > /var/lib/pbench-agent/tools-default/mpstat
+      echo "" > /var/lib/pbench-agent/tools-default/perf
+      echo "" > /var/lib/pbench-agent/tools-default/pidstat
+      master_nodes=`oc get nodes --show-labels -l pbench_agent=true,node-role.kubernetes.io/master= | grep master | awk '{print $1}'`
```
jmencak: `awk '/master/ {print $1}'`
suggest dropping grep, awk can do this for you.  Saying that, all the remaining tests have this issue.

workloads/files/workload-network-script-cm.yml
```diff
@@ -0,0 +1,961 @@
...
+      for node in $master_nodes; do
+        echo "master" > /var/lib/pbench-agent/tools-default/remote@$node
+      done
+      infra_nodes=`oc get nodes --show-labels -l pbench_agent=true,node-role.kubernetes.io/infra= | grep infra | awk '{print $1}'`
```
jmencak: `awk '/infra/ {print $1}'`

workloads/files/workload-network-script-cm.yml
```diff
@@ -0,0 +1,961 @@
...
+      for node in $infra_nodes; do
+        echo "infra" > /var/lib/pbench-agent/tools-default/remote@$node
+      done
+      worker_nodes=`oc get nodes --show-labels -l pbench_agent=true,node-role.kubernetes.io/worker= | grep worker | awk '{print $1}'`
```
jmencak: `awk '/worker/ {print $1}'`

workloads/vars/network.yml
```diff
@@ -0,0 +1,49 @@
...
+pbench_ssh_public_key_file: "{{ lookup('env', 'PBENCH_SSH_PUBLIC_KEY_FILE')|default('~/.ssh/id_rsa.pub', true) }}"
+enable_pbench_agents: "{{ lookup('env', 'ENABLE_PBENCH_AGENTS')|default(false, true)|bool }}"
+pbench_server: "{{ lookup('env', 'PBENCH_SERVER')|default('', true) }}"
+
+# Other variables for workload tests
+scale_ci_results_token: "{{ lookup('env', 'SCALE_CI_RESULTS_TOKEN')|default('', true) }}"
+job_completion_poll_attempts: "{{ lookup('env', 'JOB_COMPLETION_POLL_ATTEMPTS')|default(360, true)|int }}"
+
+# Network benchmarks specific parameters:
+network_test_uperf_image: "{{ lookup('env', 'NETWORK_TEST_UPERF_IMAGE')|default('quay.io/openshift-scale/scale-ci-uperf', true) }}"
```
jmencak, line 37: This shuffling of environment variables from to ansible variables and back from ansible variables to environment variables through config maps is very error prone in my opinion, on the other hand, at least the defaults are centralized here.  No solution is probably ideal here.

workloads/files/workload-network-script-cm.yml
```diff
@@ -0,0 +1,961 @@
...
+          if [ "${server_ready}" == "true" ]; then
+            echo "$(date -u) uperf-server-${rc_num} is ready"
+            break
+          else
+            echo "$(date -u) Waiting for uperf-server-${rc_num} to be ready"
+            sleep 1
+          fi
+          retries=$[$retries + 1]
+        done
+        [ "${server_ready}" == "false" ] && echo "$(date -u) uperf-server-${rc_num} failed to be ready on time" && exit 1
```
jmencak: Same as above.

workloads/files/workload-network-script-cm.yml
```diff
@@ -0,0 +1,961 @@
...
+          if [ "${client_ready}" == "true" ]; then
+            echo "$(date -u) uperf-client-${rc_num} is ready"
+            break
+          else
+            echo "$(date -u) Waiting for uperf-client-${rc_num} to be ready"
+            sleep 1
+          fi
+          retries=$[$retries + 1]
+        done
+        [ "${client_ready}" == "false" ] && echo "$(date -u) uperf-client-${rc_num} failed to be ready on time" && exit 1
```
jmencak: I know that `echo` will probably always exit 0, but is this guaranteed and will this always be the case? If you insist on having one-liners, I suggest:
```
[ "${client_ready}" == "false" ] && { echo "$(date -u) uperf-client-${rc_num} failed to be ready on time"; exit 1; }
```
or more readable variant:
```
[ "${client_ready}" == "false" ] && {
  echo "$(date -u) uperf-client-${rc_num} failed to be ready on time"
  exit 1
}
```
or even better something like a `die()` function?

workloads/files/workload-network-script-cm.yml
```diff
@@ -0,0 +1,961 @@
...
+      done
+
+      if [ "${NETWORK_TEST_SERVICE}" == "true" ]; then
+        clients=`oc get svc --output=json -n network0 --selector="name=client" | jq -r '[.items[].spec.clusterIP] | join(",")'`
+        servers=`oc get svc --output=json -n network0 --selector="name=server" | jq -r '[.items[].spec.clusterIP] | join(",")'`
+      else
+        clients=`oc get pods --output=json -n network0 --selector="name=client" | jq -r '[.items[].status.podIP] | join(",")'`
+        servers=`oc get pods --output=json -n network0 --selector="name=server" | jq -r '[.items[].status.podIP] | join(",")'`
+      fi
+      echo "${clients} ${servers}"
```
jmencak: Can we have better logging? The uninitiated will probably not know which is which.

workloads/files/workload-network-script-cm.yml
```diff
@@ -0,0 +1,961 @@
...
+      while [ $retries -le 60 ] ; do
+        remaining_projects=`oc get projects | egrep "^network[0-9]+ " | wc -l`
+        echo "$(date -u) Deleting network0 project..."
+        if [ $remaining_projects -eq 0 ]; then
+          break
+        fi
+        sleep 1
+        retries=$[$retries + 1]
+      done
+      [ ${remaining_projects} -gt 0 ] && echo "$(date -u) Failed to clean up network0 project" && exit 1
```
jmencak: One-liner with a potential issue as above.

workloads/files/workload-network-script-cm.yml
```diff
@@ -0,0 +1,961 @@
...
+
+      pbench-copy-results --prefix ${NETWORK_TEST_PREFIX}
+      echo "$(date -u) Completed Network benchmarks with ${pair_count} pairs"
```
jmencak: There seems to be too many of the `echo "$(date -u)"` commands, use a "log()" function to stderr for more consistency?

workloads/files/workload-network-script-cm.yml
```diff
@@ -0,0 +1,961 @@
...
+        description: A template for creating an uperf replicationcontroller
+        tags: uperf,perf
+    labels:
+      template: uperf
+    objects:
+    - apiVersion: v1
+      kind: ReplicationController
+      metadata:
+        labels:
+          name: "\${NS_VALUE}"
```
jmencak: I think we should avoid doing things like this.  When you start escaping `$`s in templates, something is wrong, probably the overall design.

workloads/files/workload-network-script-cm.yml
```diff
@@ -0,0 +1,961 @@
...
+          name: "\${ROLE}"
+          svc_selector: "\${ROLE}-\${IDENTIFIER}"
+        annotations:
+          description: Exposes and load balances the application pods
+        name: uperf-\${ROLE}-\${IDENTIFIER}
+      spec:
+        ports:
+        - name: ssh
+          port: ${NETWORK_TEST_UPERF_SSHD_PORT}
+        - name: tcp-20000
```
jmencak: Cannot we use a range of some sort?  If not, can this be templated?  So many ports copied, there must be another way.

workloads/network.yml
```diff
@@ -0,0 +1,147 @@
...
+
+    - name: Label worker nodes with pbench agent for network tests
+      shell: |
+        oc get nodes -l pbench_agent=true,node-role.kubernetes.io/worker= | grep worker | head -n 1 | awk '{print $1}' | xargs -I % oc label nodes % uperf=client
+        oc get nodes --show-labels -l pbench_agent=true,node-role.kubernetes.io/worker= | grep worker | grep -v uperf=client | head -n 1 | awk '{print $1}' | xargs -I % oc label nodes % uperf=server
+      when: enable_pbench_agents|bool
+
+    - name: Label worker nodes when there are no pbench agents for network tests
+      shell: |
+        oc get nodes -l node-role.kubernetes.io/worker= | grep worker | head -n 1 | awk '{print $1}' | xargs -I % oc label nodes % uperf=client
```
jmencak: Not important, feel free to ignore, but can be simplified with awk doing the grep.  Also suggest using `--no-headers` , but other tests have the same issue.

jmencak (commented):
Overall looking good, only nits, but runing eval on templates is not something I'd recommend and strongly suggest to change.

workloads/files/workload-network-script-cm.yml
akrzos: Ah leftover debugging, I will remove this.

workloads/files/workload-network-script-cm.yml
akrzos: Good point, I will simplify this template such that it is only an OpenShift template rather than combined env var applied template and OpenShift template.

workloads/files/workload-network-script-cm.yml
akrzos: I wish there was a range but I couldn't find any implementation of a port range but I did find this open kubernetes issue - https://github.com/kubernetes/kubernetes/issues/23864

Each pair of uperf clients needs one port per uperf "instance" thus the first pair can use 20010-20019, 2nd pair 20020-20029, ...

It actually gets worse, uperf needs a control port opened somewhere between 35000-45000(?) which is problematic for kubernetes services since each port needs to be spelled out.  I am still in the process of understanding how this control port was introduced into pbench-uperf and if we can add a mechanism to control which port it uses so we can precisely control it here.

In the meantime I'll see about moving the ports to be programmatically added.

workloads/network.yml
akrzos: Good point, I will fix this playbook and feel free to fix the others otherwise I will put together a pr to do so.

workloads/files/workload-network-script-cm.yml
jmencak: Wouldn't jinja2 for loop be of any help?

workloads/network.yml
jmencak: I'll create a PR do this.

workloads/vars/network.yml
akrzos: I agree that the shuffling will be a bit confusing at first to new comers but my opinion is the centralizing of the vars is fairly important and helpful in finding what can be changed or adjusted.  I guess it will be a trade-off we might only be able to evaluate over time.

workloads/files/workload-network-script-cm.yml
akrzos: Good idea, this actually solves the set -o pipefail issue with clusters without infra nodes.  Would you like to fix the other tests as well?  (Maybe @chaitanyaenr  would be able to post the fix today if available)

workloads/files/workload-network-script-cm.yml
akrzos: See above.  Thanks!

workloads/files/workload-network-script-cm.yml
akrzos: ditto!

workloads/files/workload-network-script-cm.yml
akrzos: Thanks I will look at the die function and/or implement the more readable function.

akrzos:
@jmencak I think I addressed most of your concerns.  I was also able to get the svc2svc benchmarks working completely and if you plan to run a short test with those, use the following uperf container image `quay.io/akrzos/scale-ci-uperf`

workloads/files/workload-network-script-cm.yml
```diff
@@ -0,0 +1,426 @@
...
+    echo "${USER_NAME:-default}:x:$(id -u):0:${USER_NAME:-default} user:${HOME}:/sbin/nologin" >> /etc/passwd
+    echo "" > /var/lib/pbench-agent/tools-default/oc
+    echo "workload" > /var/lib/pbench-agent/tools-default/label
+    if [[ -v ENABLE_PBENCH_AGENTS ]]; then
+      echo "" > /var/lib/pbench-agent/tools-default/disk
+      echo "" > /var/lib/pbench-agent/tools-default/iostat
+      echo "" > /var/lib/pbench-agent/tools-default/mpstat
+      echo "" > /var/lib/pbench-agent/tools-default/perf
+      echo "" > /var/lib/pbench-agent/tools-default/pidstat
+      master_nodes=`oc get nodes --show-labels -l pbench_agent=true,node-role.kubernetes.io/master= | awk '/master/ {print $1}'`
```
jmencak: I know I said we can drop the `grep` and move the functionality to `awk`, but is the '/master/' needed now you have a label selector ''node-role.kubernetes.io/master"?  Also, do you need `--show-labels`?

workloads/files/workload-network-script-cm.yml
```diff
@@ -0,0 +1,426 @@
...
+      for node in $master_nodes; do
+        echo "master" > /var/lib/pbench-agent/tools-default/remote@$node
+      done
+      infra_nodes=`oc get nodes --show-labels -l pbench_agent=true,node-role.kubernetes.io/infra= | awk '/infra/ {print $1}'`
```
jmencak: Same as above for master.

workloads/files/workload-network-script-cm.yml
```diff
@@ -0,0 +1,426 @@
...
+      for node in $infra_nodes; do
+        echo "infra" > /var/lib/pbench-agent/tools-default/remote@$node
+      done
+      worker_nodes=`oc get nodes --show-labels -l pbench_agent=true,node-role.kubernetes.io/worker= | awk '/worker/ {print $1}'`
```
jmencak: Same as above for master.

workloads/files/workload-network-script-cm.yml
```diff
@@ -0,0 +1,426 @@
...
+        network_yaml_template="$(cat /root/workload/s2s-network.yaml.template)"
+        cp /root/workload/svc-template.yaml /tmp/svc-template.yaml
+        # Append required ports to service template
+        for i in `seq 20000 20100`
+        do
+          echo -e "    - name: tcp-${i}\n      port: ${i}\n      protocol: TCP" >> /tmp/svc-template.yaml
+          echo -e "    - name: udp-${i}\n      port: ${i}\n      protocol: UDP" >> /tmp/svc-template.yaml
+        done
+      fi
+      eval "echo \"${network_yaml_template}\"" >> /tmp/network.yaml
```
jmencak: I'm still very uneasy about using eval for this.  It is a powerful friend, but a double-edged sword too.  In this case, you'll lose double quotes, '\\\\' will be evaluated as '\\', there is a possibility of running shell commands through the template...  There might be other side-effects I'm unaware of too.  If this is intended and you know what you're doing, fine, but personally I'd use eval 1) only for environment dumps for reference what variables were used; 2) in cases where I know exactly what I'm doing and have a limited set of input for eval. @sjug, @chaitanyaenr feel free to chime in.

jmencak (commented):
Looks much better, my only major concern is using eval.  Deferring to others for opinions.

workloads/files/workload-network-script-cm.yml
akrzos: I adjusted from `eval` to `envsubst` however we will need https://github.com/openshift-scale/images/pull/10 so the official workload container will have the command.  In the meantime you can use the one built in my namespace to test. (`quay.io/akrzos/scale-ci-workload`)

akrzos:
@jmencak @sjug @chaitanyaenr PTAL again.  I think I have addressed the major concerns.  If anyone has an advice on making anything look cleaner please speak up.  I will rerun some more versions of the test tonight to verify however I have been able to get all three major scenarios up and running now here (N2N aka P2P w/ HN, P2P, and S2S)

workloads/files/workload-network-script-cm.yml
jmencak: No doubt `envsubst` is safer than `eval`, deferring to other reviewers if they think this is an ok design.

akrzos:
I have a few more changes coming for unprivileged workload container performing the tests as well.

workloads/files/workload-network-script-cm.yml
```diff
@@ -0,0 +1,500 @@
...
+
+      if [ "${NETWORK_TEST_SERVICE}" == "true" ]; then
+        clients=`oc get svc --output=json -n network0 --selector="name=client" | jq -r '[.items[].spec.clusterIP] | join(",")'`
+        servers=`oc get svc --output=json -n network0 --selector="name=server" | jq -r '[.items[].spec.clusterIP] | join(",")'`
+      else
+        clients=`oc get pods --output=json -n network0 --selector="name=client" | jq -r '[.items[].status.podIP] | join(",")'`
+        servers=`oc get pods --output=json -n network0 --selector="name=server" | jq -r '[.items[].status.podIP] | join(",")'`
+      fi
+
+      mv ~/.ssh/config ~/.ssh/config.bak || true
```
akrzos, line 85: @jmencak Let me know if you think we should change this usage of `|| true` it allows privileged and unprivileged execution of this test.  I figured you would be best to review it.

akrzos:
@Deepthidharwar @chaitanyaenr @sjug If you all can take another look, I believe we are close to a final product at this point.  The workload pod can not be unprivileged as well as uperf pods for p2p and s2s, only the hostnetwork pods require privileged execution due to hostnetwork itself.

Let me know your thoughts so we can get this merged in.

workloads/files/workload-network-script-cm.yml
```diff
@@ -0,0 +1,500 @@
...
+            sleep 2
+          fi
+          retries=$[$retries + 1]
+        done
+        if [ "${client_ready}" == "false" ]; then
+          log "uperf-client-${rc_num} failed to be ready on time"
+          exit 1
+        fi
+        retries=0
+        while [ $retries -le 60 ] ; do
```
jmencak: Perhaps move into a function?  Code duplication of the entire client section.

README.md
```diff
@@ -337,3 +337,128 @@ NAMESPACE_CLEANUP
...
+```
+
+#### Suggested Pod2Pod HostNetwork Test Configuration
+
+```
+NETWORK_TEST_UPERF_IMAGE="quay.io/openshift-scale/scale-ci-uperf"
+NETWORK_TEST_UPERF_SSHD_PORT=20000
+NETWORK_TEST_UPERF_PORT_RANGE=""
+NETWORK_TEST_PREFIX=network
+NETWORK_TEST_PAIRS=1
```
jmencak, line 435: I know most of the other tests don't have this, but I don't understand what `NETWORK_TEST_PAIRS` is unless I look through the code.  Documentation of specific test variables would be great not only for this test, but all tests we have. In fact, documenting these vars should be compulsory.

workloads/files/workload-network-script-cm.yml
jmencak: Yeah, the joys of `set -e`.  I'd prefer to have a home directory always writable so that this doesn't fail and you don't need `|| true` or similar.  That's why I'm changing `HOME` for my tests.  Didn't run this test yet, so not sure if that's the same issue I had before I changed my `HOME`.

workloads/files/workload-network-script-cm.yml
```diff
@@ -0,0 +1,500 @@
...
+        --clients=${clients} \
+        --servers=${servers} \
+        --config=${config_prefix}_${pair_count}p_${NETWORK_TEST_INSTANCES//,/_}i_${NETWORK_TEST_PROTOCOL/,/_}
+
+      # Clean up project
+      if [[ "${NETWORK_TEST_CLEANUP}" == "true" ]]; then
+        oc delete project network0
+        retries=0
+        while [ $retries -le 60 ] ; do
+          remaining_projects=`oc get projects | egrep "^network[0-9]+ " -c || true`
```
jmencak: `--no-headers` would make things cleaner, but this is a nit.

workloads/files/workload-network-script-cm.yml
```diff
@@ -0,0 +1,500 @@
...
+        done
+        if [ ${remaining_projects} -gt 0 ]; then
+          log "Failed to clean up network0 project"
+          exit 1
+        fi
+      else
+        log "Clean up skipped"
+      fi
+
+      pbench-copy-results --prefix ${NETWORK_TEST_PREFIX}
```
jmencak: Is it guaranteed for the `NETWORK_TEST_PREFIX` not to contain spaces and special characters? Perhaps using `"` might help at least for folks trying to use spaces in the prefix.

jmencak (commented):
Only a few observations.  Didn't run any tests, will try to later on.

README.md
akrzos: In this case pairs refers to client/server uperf pairs. (It gets confusing in that instances refers to how many uperf threads run).  Also I agree in the same sense I don't understand all tests's vars and what are the values for each.  I think this should play into issue #16 though.  The most immediate solution that comes to mind is a docs folder with index and page for each workload but I would like to get some ideas from the group before paving that road.

workloads/files/workload-network-script-cm.yml
akrzos: Yeah in this case /.ssh/config does not exist and won't exist as we do not put it into the container image.  Even if we did it would be unclear what user should own it as it is the 'unprivileged' execution that hits it. Privileged execution has a $HOME of `/root` and thus find `/root/.ssh/config`.  My opinion is this is that this is an acceptable "1-line" solution.  Another option would be just writing an if statement however that would add a few more lines to each of the two statements that need `|| true`  I like to think that this is an acceptable hack considering it is not complex and that it does the job as needed with less lines.

workloads/files/workload-network-script-cm.yml
akrzos: Got it, moved this and the log function into a separate functions.sh file sourced at the top.

workloads/files/workload-network-script-cm.yml
akrzos: Thanks, added quotes.

workloads/files/workload-network-script-cm.yml
akrzos: NP thanks!

jmencak:
I've tested the `P2P` scenario and it worked out of the box.  Overall, it was a positive experience, even thought the playbook itself actually failed and I didn't let the test itself finish.  Timeout waiting for job completion after 111m57.127s -- perhaps bump up default `job_completion_poll_attempts` for suggested configs.  The test itself was still running after 3h51m having collected 19GB of data.  I've noticed quite a few `*-fail<N>` pbench_uperf directories which started to appear after the first `udp_rr-` test.  I've wiped the cluster after seeing `24-udp_rr-16384B-2i-fail6` which was presumably supposed to be the last iteration.  I love the "Suggested XYZ Test Configuration" in the readme, it should be compulsory for *every* workload, especially more complex ones.  A typical duration of tests running with a suggested configuration would help users too, as this was the first time for me running the network tests and I didn't know what to expect and whether they were running correctly.  Logs and rsh into the `scale-ci-network-xyz` container helped though.  Every long running test should have a `<TEST_NAME>_SMOKE_TEST` option to establish basic test functionality.  You don't want to find out after having run your test for 3 hours you haven't set your `PBENCH_SERVER` or similar properly.  I also believe documentation for workloads show probably tell the users which namespaces the workload creates.  No workload has sufficient documentation at this point.  One last thing I'd recommend is to turn `network0` namespace into a variable and adjust the associated code.

PR #20, opened 2019-06-12
Add deployments per ns cluster limits workload

chaitanyaenr:
This workload creates desired deployments per ns to test the current
OCP limit as well as push the limit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -466,3 +466,52 @@ NETWORK_TEST_INSTANCES=1,2
 NETWORK_TEST_SAMPLES=3
 NETWORK_TEST_CLEANUP=true
 ```
+
+## Deployments per namespace cluster limits Workload
+
+The Deployments per namespace cluster limits workload playbook is `workloads/cluster-limits-deployments-per-ns.yml` and will run the Deployments per namespace cluster limits test on your RHCOS cluster.
+
+Running from CLI:
+
+```sh
+$ cp workloads/inventory.example inventory
+$ # Add orchestration host to inventory
+$ # Edit vars in workloads/vars/deployments-per-ns.yml or define Environment vars (See below)
+$ time ansible-playbook -vv -i inventory workloads/cluster-limits-deployments-per-ns.yml
+```
+
+### Environment variables for Deployments per namespace cluster limits workload playbook
+
+```
+###############################################################################
+# Ansible SSH variables.
+###############################################################################
+PUBLIC_KEY
+PRIVATE_KEY
+ORCHESTRATION_USER
+
+###############################################################################
+# RHCOS workload variables.
+###############################################################################
+WORKLOAD_IMAGE
+WORKLOAD_JOB_NODE_SELECTOR
+WORKLOAD_JOB_TAINT
+WORKLOAD_JOB_PRIVILEGED
+
+KUBECONFIG_FILE
+
+PBENCH_SSH_PRIVATE_KEY_FILE
+PBENCH_SSH_PUBLIC_KEY_FILE
+ENABLE_PBENCH_AGENTS
+PBENCH_SERVER
+
+SCALE_CI_RESULTS_TOKEN
+JOB_COMPLETION_POLL_ATTEMPTS
+
+# Deployments per namespace cluster limits workload specific parameters:
+DEPLOYMENTS_PER_NS_TEST_PREFIX
+DEPLOYMENTS_PER_NS_CLEANUP
+DEPLOYMENTS_PER_NS_BASENAME
+DEPLOYMENTS_PER_NS_COUNT
+DEPLOYMENTS_PER_NS_POD_IMAGE
+```
--- a/workloads/baseline.yml
+++ b/workloads/baseline.yml
@@ -58,7 +58,7 @@
 
     - name: Check if scale-ci-tooling namespace exists
       shell: |
-        oc get projects | grep "scale-ci-tooling"
+        oc get project scale-ci-tooling
       ignore_errors: true
       changed_when: false
       register: scale_ci_tooling_ns_exists
--- a/workloads/cluster-limits-deployments-per-ns.yml
+++ b/workloads/cluster-limits-deployments-per-ns.yml
@@ -1,16 +1,16 @@
 ---
 #
-# Runs MasterVertical on an existing RHCOS cluster
+# Runs deployments-per-ns cluster limits test on an existing RHCOS cluster
 #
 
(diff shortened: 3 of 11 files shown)


workloads/cluster-limits-deployments-per-ns.yml
```diff
@@ -0,0 +1,125 @@
...
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/workload-deployments-per-ns-script-cm.yml"
+
+    - name: Check if scale-ci-tooling namespace exists
+      shell: |
+        oc get projects | grep "scale-ci-tooling"
+      ignore_errors: true
+      changed_when: false
+      register: scale_ci_tooling_ns_exists
+
+    - name: Ensure any stale scale-ci-podvertical job is deleted
```
chaitanyaenr: s/scale-ci-podvertical/deployments-per-ns

README.md
```diff
@@ -337,3 +337,53 @@ NAMESPACE_CLEANUP
 HTTP_STRESS_CONTAINER_IMAGE
 HTTP_SERVER_CONTAINER_IMAGE
 ```
+
+## Deployments per namespace cluster limits Workload
+
+The Deployments per namespace cluster limits workload playbook is `workloads/cluster-limits-deployments-per-ns.yml` and will run the Deployments per namespace cluster limits test on your RHCOS cluster.
```
akrzos, line 472: `cluster-limits-deployments-per-ns.yml` are we sure we want to name the playbook for this test with a  prefix of cluster limits?  I am not for/against it just curious if we think we should prefix all cluster limit playbooks as such.

workloads/templates/workload-deployments-per-ns-script-cm.yml.j2
```diff
@@ -0,0 +1,125 @@
...
+        strategy:
+          type: Rolling
+    parameters:
+    - name: IDENTIFIER
+      description: Number to append to the name of resources
+      value: '1'
+      required: true
+    - name: IMAGE
+      description: Image to use for deploymentConfig
+      value: {{deployments_per_ns_pod_image}}
```
akrzos: Instead of feeding the image in here, perhaps put it in the cluster loader yaml as a parameter so cluster loader renders this template with the image.  That way this template becomes "static" content rather than a rendered/templated content.

workloads/templates/workload-deployments-per-ns-script-cm.yml.j2
```diff
@@ -0,0 +1,125 @@
...
+      metadata:
+        name: deploymentconfig${IDENTIFIER}
+      spec:
+        template:
+          metadata:
+            labels:
+              name: replicationcontroller${IDENTIFIER}
+          spec:
+            containers:
+            - name: pause${IDENTIFIER}
```
akrzos: Since these pods may not be pause pod containers we might want to have a different name perhaps "dpns" or deployments per namespace shortened?  Also we shouldn't need the ${IDENTIFIER} on the container name, just on the pod/rc names.

workloads/cluster-limits-deployments-per-ns.yml
```diff
@@ -0,0 +1,125 @@
...
+    - name: Check if scale-ci-tooling namespace exists
+      shell: |
+        oc get projects | grep "scale-ci-tooling"
+      ignore_errors: true
+      changed_when: false
+      register: scale_ci_tooling_ns_exists
+
+    - name: Ensure any stale scale-ci-deployments-per-namespace job is deleted
+      shell: |
+        oc delete job scale-ci-deployments-per-namespace -n scale-ci-tooling
```
akrzos: This `scale-ci-deployments-per-namespace` should be `scale-ci-deployments-per-ns`.

akrzos (changes requested):
It run successfully for me but I found one small issue with the "stale" job delete task in the playbook.  I also suggested feeding the template-ed image in to the cluster loader config file since there are already some ansible vars template-ed in there and we can keep the deployment config template static then.

README.md
chaitanyaenr: I think it would be better to have a cluster limits prefix for the limits workloads for them to be identified easily but It's just a thought, we can remove the prefix as well.

workloads/cluster-limits-deployments-per-ns.yml
```diff
@@ -0,0 +1,125 @@
...
+        - src: kubeconfig-secret.yml.j2
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/kubeconfig-secret.yml"
+        - src: workload-job.yml.j2
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/workload-job.yml"
+        - src: workload-deployments-per-ns-script-cm.yml.j2
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/workload-deployments-per-ns-script-cm.yml"
+
+    - name: Check if scale-ci-tooling namespace exists
+      shell: |
+        oc get projects | grep "scale-ci-tooling"
```
jmencak: Would `oc get project scale-ci-tooling` work?

chaitanyaenr: Yes, it will work. Changed it in all the workload playbooks. Thanks.

chaitanyaenr:
Will rebase to resolve the conflicts.

chaitanyaenr:
@akrzos @jmencak made the changes to address the comments. PTAL.

workloads/baseline.yml
```diff
@@ -58,7 +58,7 @@
 
     - name: Check if scale-ci-tooling namespace exists
       shell: |
-        oc get projects | grep "scale-ci-tooling"
+         oc get project scale-ci-tooling
```
akrzos: nitpick but extra space in front of the `oc` command here.  Each indent should be 2 spaces

workloads/http.yml
```diff
@@ -63,7 +63,7 @@
 
     - name: Check if scale-ci-tooling namespace exists
       shell: |
-        oc get projects | grep "scale-ci-tooling"
+         oc get project scale-ci-tooling
```
akrzos: Looks like same nitpick here for all of these.

akrzos (changes requested):
Small nitpick on spacing, I also believe we need to modify the network playbook for checking if the scale-ci-tooling namespace exists.

chaitanyaenr:
@akrzos fixed it. PTAL.

akrzos (approved):
LGTM

PR #24, opened 2019-06-14
PVC scale test

ekuric:
PVC scale test: 

- creates pods 
- attaches pvc 
- collect timestamps for pods

### DIFF
--- a/workloads/pvcscale.yml
+++ b/workloads/pvcscale.yml
@@ -1,0 +1,137 @@
+---
+#
+# Runs PVC Scale test on an existing RHCOS cluster
+#
+
+- name: Runs PVC scale on a RHCOS cluster
+  hosts: orchestration
+  gather_facts: true
+  remote_user: "{{orchestration_user}}"
+  vars_files:
+    - vars/pvcscale.yml
+  vars:
+    workload_job: "pvcscale"
+  tasks:
+    - name: Create scale-ci-tooling directory
+      file:
+        path: "{{ansible_user_dir}}/scale-ci-tooling"
+        state: directory
+
+    - name: Copy workload files
+      copy:
+        src: "{{item.src}}"
+        dest: "{{item.dest}}"
+      with_items:
+        - src: scale-ci-tooling-ns.yml
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/scale-ci-tooling-ns.yml"
+
+    - name: Slurp kubeconfig file
+      slurp:
+        src: "{{kubeconfig_file}}"
+      register: kubeconfig_file_slurp
+
+    - name: Slurp ssh private key file
+      slurp:
+        src: "{{pbench_ssh_private_key_file}}"
+      register: pbench_ssh_private_key_file_slurp
+
+    - name: Slurp ssh public key file
+      slurp:
+        src: "{{pbench_ssh_public_key_file}}"
+      register: pbench_ssh_public_key_file_slurp
+
+    - name: Template workload templates
+      template:
+        src: "{{item.src}}"
+        dest: "{{item.dest}}"
+      with_items:
+        - src: pbench-cm.yml.j2
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/pbench-cm.yml"
+        - src: pbench-ssh-secret.yml.j2
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/pbench-ssh-secret.yml"
+        - src: kubeconfig-secret.yml.j2
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/kubeconfig-secret.yml"
+        - src: workload-job.yml.j2
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/workload-job.yml"
+        - src: workload-pvcscale-script-cm.yml.j2
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/workload-pvcscale-script-cm.yml"
+          #- src: pvcrole.yml.j2
+          #dest: "{{ansible_user_dir}}/scale-ci-tooling/pvcrole.yml"
+          #- src: pvcrolebinding.yml.j2
+          #dest: "{{ansible_user_dir}}/scale-ci-tooling/pvcrolebinding.yml"
+
+    - name: Check if scale-ci-tooling namespace exists
+      shell: |
+        oc get projects | grep "scale-ci-tooling"
+      ignore_errors: true
+      changed_when: false
+      register: scale_ci_tooling_ns_exists
+    - name: print status of scale_ci_tooling_ns_exists
+      debug:
+        var: scale_ci_tooling_ns_exists
+
+    - name: Ensure any stale scale-ci-pvcscale job is deleted
+      shell: |
+        oc delete job scale-ci-pvcscale -n scale-ci-tooling
+      register: scale_ci_tooling_project
+      failed_when: scale_ci_tooling_project.rc == 0
+      until: scale_ci_tooling_project.rc == 1
+      retries: 60
+      delay: 1
+      when: scale_ci_tooling_ns_exists.rc == 0
+
+    - name: Block for non-existing tooling namespace
+      block:
+        - name: Create tooling namespace
+          shell: |
+            oc create -f {{ansible_user_dir}}/scale-ci-tooling/scale-ci-tooling-ns.yml
+      when: scale_ci_tooling_ns_exists.rc != 0
+
+    - name: Create tooling service account
+      shell: |
+        oc create serviceaccount useroot -n scale-ci-tooling
+        oc adm policy add-scc-to-user privileged -z useroot -n scale-ci-tooling
+      when: enable_pbench_agents|bool
+
+    - name: Add serviceaccount for useroot when pbench is not used
+      shell: |
+        oc create serviceaccount useroot -n scale-ci-tooling
+      when: not enable_pbench_agents|bool
+
+    - name: Create/replace kubeconfig secret
+      shell: |
+        oc replace --force -n scale-ci-tooling -f "{{ansible_user_dir}}/scale-ci-tooling/kubeconfig-secret.yml"
+
+    - name: Create/replace the pbench configmap
+      shell: |
+        oc replace --force -n scale-ci-tooling -f "{{ansible_user_dir}}/scale-ci-tooling/pbench-cm.yml"
+
+    - name: Create/replace pbench ssh secret
+      shell: |
+        oc replace --force -n scale-ci-tooling -f "{{ansible_user_dir}}/scale-ci-tooling/pbench-ssh-secret.yml"
+
+    - name: Create/replace workload script configmap
+      shell: |
+        oc replace --force -n scale-ci-tooling -f "{{ansible_user_dir}}/scale-ci-tooling/workload-pvcscale-script-cm.yml"
+
+    - name: Create/replace workload job to that runs workload script
+      shell: |
+        oc replace --force -n scale-ci-tooling -f "{{ansible_user_dir}}/scale-ci-tooling/workload-job.yml"
+
+    - name: Poll until job is active
+      shell: |
+        oc get job scale-ci-pvcscale -n scale-ci-tooling -o json
+      register: job_json
+      retries: 60
+      delay: 2
+      until: job_json.stdout | from_json | json_query('status.active==`1`')
+
+    - name: Poll until job is complete
+      shell: |
+        oc get job scale-ci-pvcscale -n scale-ci-tooling -o json
+      register: job_json
+      retries: "{{job_completion_poll_attempts}}"
+      delay: 10
+      until: job_json.stdout | from_json | json_query('status.succeeded==`1` || status.failed==`1`')
+      failed_when: job_json.stdout | from_json | json_query('status.succeeded==`1`') == false
+      when: job_completion_poll_attempts|int > 0
--- a/workloads/templates/workload-pvcscale-script-cm.yml.j2
+++ b/workloads/templates/workload-pvcscale-script-cm.yml.j2
@@ -1,0 +1,354 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: scale-ci-workload-script
+data:
+  run.sh: |
+    #!/bin/sh
+    set -eo pipefail
+    # pbench Configuration
+    echo "$(date -u) Configuring pbench for PVC scale test"
+    mkdir -p /var/lib/pbench-agent/tools-default/
+    echo "${USER_NAME:-default}:x:$(id -u):0:${USER_NAME:-default} user:${HOME}:/sbin/nologin" >> /etc/passwd
+    echo "" > /var/lib/pbench-agent/tools-default/oc
+    echo "workload" > /var/lib/pbench-agent/tools-default/label
+    if [[ -v ENABLE_PBENCH_AGENTS ]]; then
+      echo "" > /var/lib/pbench-agent/tools-default/disk
+      echo "" > /var/lib/pbench-agent/tools-default/iostat
+      echo "" > /var/lib/pbench-agent/tools-default/mpstat
+      echo "" > /var/lib/pbench-agent/tools-default/perf
+      echo "" > /var/lib/pbench-agent/tools-default/pidstat
+      master_nodes=`oc get nodes -l pbench_agent=true,node-role.kubernetes.io/master= --no-headers | awk '{print $1}'`
+      for node in $master_nodes; do
+        echo "master" > /var/lib/pbench-agent/tools-default/remote@$node
+      done
+      infra_nodes=`oc get nodes -l pbench_agent=true,node-role.kubernetes.io/infra= --no-headers | awk '{print $1}'`
+      for node in $infra_nodes; do
+        echo "infra" > /var/lib/pbench-agent/tools-default/remote@$node
+      done
+      worker_nodes=`oc get nodes -l pbench_agent=true,node-role.kubernetes.io/worker= --no-headers | awk '{print $1}'`
+      for node in $worker_nodes; do
+        echo "worker" > /var/lib/pbench-agent/tools-default/remote@$node
+      done
+    fi
+    source /opt/pbench-agent/profile
+    echo "$(date -u) Done configuring pbench for PVC scale test"
+    # End pbench Configuration
+
+    # Test Configuration
+    echo "$(date -u) Running PVC scale test"
+    pbench-user-benchmark --config="{{ pvcscale_test_prefix }}-pods-{{ pvcscale_maxpods }}-sc-{{ pvcscale_storageclass }}-create_pods" -- 'VIPERCONFIG=/root/workload/pvcscale.yml openshift-tests run-test "[Feature:Performance][Serial][Slow] Load cluster should load the cluster [Suite:openshift]"'
+
+    echo "$(date -u) Pods/PVC are crated ..."
+    # End Test Configuration
+    echo "Running pvc data collection"
+
+    # wait until all pods are started and then collect data
+    while [[ $(oc get pods -n {{ pvcscale_basename }}0 | grep pvc | grep Run |wc -l) -lt {{ pvcscale_maxpods }}  ]] ; do
+      sleep 30
+      echo "Waiting on pods to start..."
+    done
+    cp /root/workload/result.sh /tmp/
+    # todo: define how long to wait before stopping in case pods are not starting
+    pbench-user-benchmark --config="{{ pvcscale_test_prefix }}-pods-{{ pvcscale_maxpods }}-sc-{{ pvcscale_storageclass }}" -- /tmp/result.sh
+    pbench-copy-results --prefix "{{ pvcscale_test_prefix }}{{ pvcscale_maxpods }}{{ pvcscale_storageclass }}"
+    if [[ {{ pvcscale_cleanup }} == "true" ]]; then
+       oc delete project {{ pvcscale_basename }}0
+       while [ "$(oc get project |grep {{ pvcscale_basename }}0 | awk '{print $1}')" == {{ pvcscale_basename }}0 ]; do
+         echo "waiting on project {{ pvcscale_basename }}0 to disappear ..."
+         sleep 10
+       done
+       echo "Project {{ pvcscale_basename }}0 is deleted ... test finished"
+    elif [[ {{ pvcscale_cleanup }} == "false" ]]; then
+       echo "Test is done, but project {{ pvcscale_basename }}0 is not be deleted due to PVCSCALE_CLAENUP=false"
+    fi
+  result.sh: |
+    #!/bin/bash
+    cp /root/.kube/config /tmp/config
+    export KUBECONFIG=/tmp/config
+    oc create rolebinding pvcrolepvscale --clusterrole=view --serviceaccount=scale-ci-tooling:useroot --namespace={{ pvcscale_basename }}0
+    KUBEHOST=$(oc cluster-info | head -1 | cut -d'/' -f3 | cut -d':' -f1)
+    TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
+    oc login --token=$TOKEN https://$KUBEHOST:6443 --certificate-authority=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
+    python /root/workload/check_pvpods.py --host $KUBEHOST --port 6443 --to $TOKEN -n {{ pvcscale_basename }}0 --check_pods --resdir "${benchmark_results_dir}"/pvcscaletest
+    cat "${benchmark_results_dir}"/pvcscaletest/pods_ready{{ pvcscale_basename }}0.csv  | cut -d',' -f9 | cut -d'T' -f2|cut -d'Z' -f1 >> "${benchmark_results_dir}"/pvcscaletest/pvcpod.csv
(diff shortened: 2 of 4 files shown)


ekuric:
Test [log](https://gist.github.com/ekuric/17de3991a3983f4a1f9095adc21d6b7b)

workloads/pvcscale.yml
```diff
@@ -0,0 +1,141 @@
...
+        - src: kubeconfig-secret.yml.j2
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/kubeconfig-secret.yml"
+        - src: workload-job.yml.j2
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/workload-job.yml"
+        - src: workload-pvcscale-script-cm.yml.j2
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/workload-pvcscale-script-cm.yml"
+
+    - name: Check if scale-ci-tooling namespace exists
+      shell: |
+        oc get projects | grep "scale-ci-tooling"
```
jmencak, line 65: I suggest using `oc get project scale-ci-tooling` instead.

workloads/storage/pvcscale_README.md
```diff
@@ -0,0 +1,116 @@
+### Purpose
+PVC scale test is designed to create desired number of pods and attach PVC to every pod from storage backend dynamically 
+using storage storage class. 
+
+### Prerequesties
+- PVC test requieres functional storage backend and storage class.
```
jmencak: requires

workloads/storage/pvcscale_README.md
```diff
@@ -0,0 +1,116 @@
...
+- PVC test requieres functional storage backend and storage class. 
+- Pbench [web server](https://distributed-system-analysis.github.io/pbench/doc/server/pbench-web-server.html)
+
+### Environment variables for PVC Scale workload playbook
+
+```
+###############################################################################
+# Ansible SSH variables.
+###############################################################################
+export PUBLIC_KEY
```
jmencak: I'd prefer to drop all of the exports even though they're necessary.  Folks should know this and they're not interested in `export` keyword, they're interested in the variable itself.

workloads/storage/pvcscale_README.md
```diff
@@ -0,0 +1,116 @@
...
+
+ORCHESTRATION_USER="root"
+###############################################################################
+# RHCOS workload variables.
+###############################################################################
+export WORKLOAD_IMAGE="quay.io/openshift-scale/scale-ci-workload"
+export KUBECONFIG_FILE="path to kube config file" 
+export PBENCH_SSH_PRIVATE_KEY_FILE
+export PBENCH_SSH_PUBLIC_KEY_FILE
+export ENABLE_PBENCH_AGENTS=true
```
jmencak: Is `=true` intentional?

workloads/storage/pvcscale_README.md
```diff
@@ -0,0 +1,116 @@
...
+export PBENCH_SERVER="fqdn of pbench web server" 
+
+export SCALE_CI_RESULTS_TOKEN
+export JOB_COMPLETION_POLL_ATTEMPTS
+
+# PVC Scale workload specific parameters:
+export PVCSCALE_TEST_PREFIX="test prefix"
```
jmencak: Great you're trying to document the vars.  I'd keep the docs for the vars consistent across the workloads though.  Granted, there's not much of a documentation for these, but shouldn't we use standard comments itself, like `# comment`?

workloads/storage/pvcscale_README.md
```diff
@@ -0,0 +1,116 @@
...
+export PVCSCALE_POD_IMAGE="Image which will be used to start pods"
+export PVCSCALE_STEPSIZE="how many Pods to create in one step"
+export PVCSCALE_PAUSE="how long to wait between steps"
+export PVCSCALE_STORAGE_SIZE="pvc size - eg 1Gi, 2Gi"
+export PVCSCALE_STORAGECLASS="storageclass_name"
+export ACCESS_MODES=ReadWriteOnce|ReadWriteMany|ReadOnlyMany
+export PVC_SCALE_TEST=true|false
+``` 
+
+Example paramters are listed below
```
jmencak: parameters

workloads/templates/workload-pvcscale-script-cm.yml.j2
```diff
@@ -0,0 +1,367 @@
...
+      echo "" > /var/lib/pbench-agent/tools-default/disk
+      echo "" > /var/lib/pbench-agent/tools-default/iostat
+      echo "" > /var/lib/pbench-agent/tools-default/mpstat
+      echo "" > /var/lib/pbench-agent/tools-default/perf
+      echo "" > /var/lib/pbench-agent/tools-default/pidstat
+      master_nodes=`oc get nodes -l pbench_agent=true,node-role.kubernetes.io/master= --no-headers | awk '{print $1}'`
+      for node in $master_nodes; do
+        echo "master" > /var/lib/pbench-agent/tools-default/remote@$node
+      done
+      infra_nodes=`oc get nodes --show-labels -l pbench_agent=true,node-role.kubernetes.io/infra= | grep infra | awk '{print $1}'`
```
jmencak: Is the `grep infra` necessary here?  Please see the `master_nodes` above.

workloads/templates/workload-pvcscale-script-cm.yml.j2
```diff
@@ -0,0 +1,367 @@
...
+      for node in $infra_nodes; do
+        echo "infra" > /var/lib/pbench-agent/tools-default/remote@$node
+      done
+      worker_nodes=`oc get nodes --show-labels -l pbench_agent=true,node-role.kubernetes.io/worker= | grep worker | awk '{print $1}'`
```
jmencak: Is the `grep worker` necessary here?  Please see the `master_nodes` above.

workloads/templates/workload-pvcscale-script-cm.yml.j2
```diff
@@ -0,0 +1,367 @@
...
+    echo "$(date -u) Completed PVC scale test"
+    # End Test Configuration
+    echo "Running pvc data collection"
+
+    # remove once Alex get python-requests into image
+    yum install python-requests -y
+    oc create -f /root/workload/pvcrole.yml
+    oc create -f /root/workload/pvcrolebinding.yml
+    # wait until all pods are started and then collect data
+    while [[ $(oc get pods -n {{ pvcscale_basename }}0 | grep pvc | grep Run |wc -l) -lt {{ pvcscale_maxpods }}  ]] ; do sleep 30; echo "Waiting on pods to start..."; done
```
jmencak: Formatting.  Also, could the multiple `grep`s be replaced by something safer?

workloads/templates/workload-pvcscale-script-cm.yml.j2
```diff
@@ -0,0 +1,367 @@
...
+        api_pods = '%s/namespaces/%s/pods' % (self.base_api, self.namespace)
+        parsed_json = self.get_json(api_pods)
+        pods = []
+
+        # sort pods based on creationTime
+        for item in parsed_json["items"]:
+            pods.append(item)
+            pods = sorted(pods, key=lambda k: k["status"]["containerStatuses"][0]["state"]["running"]["startedAt"],reverse=False)
+            with open(self.resdir+"/""pods_"+str(ns)+".json", "w") as allpods:
+                json.dump(pods,allpods, indent=4)
```
jmencak: `, allpods`?

workloads/templates/workload-pvcscale-script-cm.yml.j2
```diff
@@ -0,0 +1,367 @@
...
+    - name: PVC_NAME
+      description: PVC name
+      required: true
+      from: pvc[a-z0-9]{10}
+      generate: expression
+    - name: IDENTIFIER
+      description: Number to append to the name of resources
+      value: '1'
+  pvcrole.yml: |
+   ---
```
jmencak: Are `---` necessary?

jmencak (commented):
Only initial comments -- this is a very quick review pass.  I got lost in the Python code. :) Thanks for the PR, Elko!

workloads/templates/workload-pvcscale-script-cm.yml.j2
```diff
@@ -0,0 +1,367 @@
...
+    # Test Configuration
+    echo "$(date -u) Running PVC scale test"
+    pbench-user-benchmark --config="{{ pvcscale_test_prefix }}-pods-{{ pvcscale_maxpods }}-sc-{{ pvcscale_storageclass }}-create_pods" -- 'VIPERCONFIG=/root/workload/pvcscale.yml openshift-tests run-test "[Feature:Performance][Serial][Slow] Load cluster should load the cluster [Suite:openshift]"'
+
+    echo "$(date -u) Completed PVC scale test"
+    # End Test Configuration
+    echo "Running pvc data collection"
+
+    # remove once Alex get python-requests into image
+    yum install python-requests -y
```
akrzos: Lets open a pr to install `python-requests` into the workload container image.  You can simply add it to this yum statement [here](https://github.com/openshift-scale/images/blob/master/scale-ci-workload/Dockerfile#L24) and then open that pr and we will get it into the image.

workloads/storage/pvcscale_README.md
```diff
@@ -0,0 +1,116 @@
+### Purpose
```
akrzos: The extra docs here are great.  We do have an open issue #16 for discussing on how to organize the docs and so far we think we will create a `docs` directory with an index and links to each workload's specific docs page.  I think this would be a great first example of that (if we don't start building it shortly)  Could we move this to `docs/pvcscale.md`?

workloads/templates/workload-pvcscale-script-cm.yml.j2
```diff
@@ -0,0 +1,367 @@
...
+      annotations:
+        descriptions: PVC scale test - start pods and mount PVC from StorageClass
+        tags: pvc, pvc-test
+    objects:
+    - apiVersion: v1
+      kind: PersistentVolumeClaim
+      metadata:
+        name: "${PVC_NAME}"
+        annotations:
+          volume.beta.kubernetes.io/storage-class: {{ pvcscale_storageclass }}
```
akrzos, line 313: My suggestion here would be to convert this entirely to an OpenShift template and have the ansible templated vars rendered in the cluster loader configuration.  That would allow this template to be easier to validate for the future.  An example of this would be what @chaitanyaenr just did for the deployments per namespace workload.  Example in the openshift template - https://github.com/openshift-scale/workloads/blob/master/workloads/templates/workload-deployments-per-ns-script-cm.yml.j2#L116-L119 and the referencing cluster loader file with the ansible var: https://github.com/openshift-scale/workloads/blob/master/workloads/templates/workload-deployments-per-ns-script-cm.yml.j2#L58

Looks like a few vars here just need adjustment to align with that method of vars/templating.

workloads/vars/pvcscale.yml
```diff
@@ -0,0 +1,48 @@
...
+pvcscale_basename: "{{ lookup('env', 'PVCSCALE_BASENAME')|default('pvcscale', true) }}"
+pvcscale_maxpods: "{{ lookup('env', 'PVCSCALE_MAXPODS')|default(10, true)|int }}"
+pvcscale_pod_image: "{{ lookup('env', 'PVCSCALE_POD_IMAGE')|default('quay.io/openshift-scale/scale-ci-workload', true) }}"
+pvcscale_stepsize: "{{ lookup('env', 'PVCSCALE_STEPSIZE')|default(1, true)|int }}"
+pvcscale_pause: "{{ lookup('env', 'PVCSCALE_PAUSE')|default(0, true)|int }}"
+pvcscale_storage_size: "{{ lookup('env', 'PVCSCALE_STORAGE_SIZE')|default('1Gi',true) }}"
+pvcscale_storageclass: "{{ lookup('env', 'PVCSCALE_STORAGECLASS')|default('csi-rbd', true) }}"
+access_modes: "{{ lookup('env', 'ACCESS_MODES')|default('ReadWriteOnce',true) }}"
+pvc_name: "{{ lookup('env', 'PVC_NAME')|default('pvctest',true) }}"
+pvc_scale_test: "{{ lookup('env', 'PVC_SCALE_TEST')|default('pvc_scale_test',true) }}"
```
akrzos: I think we can eliminate this var `pvc_scale_test` and if we can't shouldn't its default be `true`?

workloads/pvcscale.yml
```diff
@@ -0,0 +1,141 @@
...
+          when: enable_pbench_agents|bool
+      when: scale_ci_tooling_ns_exists.rc != 0
+
+    - name: PVC test specific user configuration 
+      block:
+        - name: Add htpasswd identity provider for user when running pvc scale test
+          shell: |
+            htpasswd -c -B -b /tmp/htpasswd useroot useroot
+            oc create secret generic htpass-secret --from-file=/tmp/htpasswd -n openshift-config
+      when: pvc_scale_test|bool
```
akrzos: Since this is the `pvcscale.yml` playbook which means the user intends to run the pvcscale workload, can we eliminate this var such that these tasks always execute?

workloads/pvcscale.yml
```diff
@@ -0,0 +1,141 @@
...
+            oc adm policy add-scc-to-user privileged -z useroot -n scale-ci-tooling
+          when: enable_pbench_agents|bool
+      when: scale_ci_tooling_ns_exists.rc != 0
+
+    - name: PVC test specific user configuration 
+      block:
+        - name: Add htpasswd identity provider for user when running pvc scale test
+          shell: |
+            htpasswd -c -B -b /tmp/htpasswd useroot useroot
+            oc create secret generic htpass-secret --from-file=/tmp/htpasswd -n openshift-config
```
akrzos: While running/rerunning this I ran into this error related to the secret created here:
```
FAILED! => {"changed": true, "cmd": "htpasswd -c -B -b /tmp/htpasswd useroot useroot\noc create secret generic htpass-secret --from-file=/tmp/htpasswd -n openshift-config\n", "delta": "0:00:00.379115", "end": "2019-06-14 22:19:38.052228", "msg": "non-zero return code", "rc": 1, "start": "2019-06-14 22:19:37.673113", "stderr": "Adding password for user useroot\nError from server (AlreadyExists): secrets \"htpass-secret\" already exists", "stderr_lines": ["Adding password for user useroot", "Error from server (AlreadyExists): secrets \"htpass-secret\" already exists"], "stdout": "", "stdout_lines": []}
```

We should make this a bit more robust incase the secret already exists.

workloads/templates/workload-pvcscale-script-cm.yml.j2
```diff
@@ -0,0 +1,367 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: scale-ci-workload-script
+data:
+  run.sh: |
+    #!/bin/sh
+    # pbench Configuration
```
akrzos, line 9: As a best practice we should have `set -eo pipefail` up here.

akrzos (changes requested):
First off, Thanks for the workload!! I was able to get the cluster loader portion to run however I couldn't get the full results.  I put in some feedback throughout the commit and we can work together next week at the F2F to help get this all up and running.  Thanks and see you soon!

workloads/pvcscale.yml
ekuric: I agree,I think we have change this test wide for all ( if not already )

workloads/storage/pvcscale_README.md
ekuric: ack!

workloads/storage/pvcscale_README.md
ekuric: can be, but better to stay in line with other parameters ( no value )

workloads/storage/pvcscale_README.md
ekuric: ok, will change and we will have to document for other tests parameters too

workloads/templates/workload-pvcscale-script-cm.yml.j2
ekuric: Ok, I will open this PR.

workloads/storage/pvcscale_README.md
ekuric: will move it to docs/pvscale.md

workloads/templates/workload-pvcscale-script-cm.yml.j2
ekuric: I have to test this, PVC names need to be different for every pod, I did not tested this

workloads/vars/pvcscale.yml
ekuric: we can but I added it there for case in future it all is in one place / playbook

workloads/pvcscale.yml
ekuric: nothing created it,  so it should not be there, if it there it will be used,

workloads/templates/workload-pvcscale-script-cm.yml.j2
ekuric: we have in all scripts `pbench configuration` and that part is same , easy to fix it here, but I would like the most to have this part in single file valid for all tests

akrzos: We could refactor at some point the pbench configuration part of the run script.  For now I think it's simple enough to remain a boiler plate code but I'm open to ideas.  As mentioned I think all scripts should run with the above statement though.

workloads/pvcscale.yml
akrzos: However the issue is I wanted to run more than once, the task should permit "rerunning".

ekuric: it should delete secret at end of test https://github.com/openshift-scale/workloads/blob/b6e7b19d0c242d8e0e89a8e033499460a88c40f4/workloads/pvcscale.yml#L138 so next run will create it again, but I will experiment a bit more and see what is best

workloads/storage/pvcscale_README.md
akrzos: #29 Will add the docs directory and should provide examples to what/how to document the workload like the rest.  (If you get a chance please review it)

workloads/storage/pvcscale_README.md
ekuric: I will drop README from this PR and create new one on top of master per #16

workloads/storage/pvcscale_README.md
ekuric: will drop pvcscale_README.md from this PR and create new as we are discussing that in #16

workloads/pvcscale.yml
ekuric: ok, removing pvc_scale_test

workloads/pvcscale.yml
ekuric: added ` oc delete secret -n openshift-config htpass-secret` prior create step,this will ensure secret is created always (in case test did nto reach till end where it is anyway deleted ). I am investigating better way to handle this,will created new PR for this.

workloads/templates/workload-pvcscale-script-cm.yml.j2
ekuric: added `set -eo pipefail`

workloads/templates/workload-pvcscale-script-cm.yml.j2
ekuric: yes, json.dump stuff,

ekuric:
@akrzos @jmencak I removed documentation part , will add via new PR, other than that I tested this and it runs also in unprivileged mode, @akrzos if you can test it, it worked fine with these [vars](https://gist.github.com/ekuric/02971d2245d001feffca0567e29f5cd6)

workloads/templates/workload-pvcscale-script-cm.yml.j2
```diff
@@ -0,0 +1,367 @@
...
+
+      if args.check_pods:
+        myos.get_pods()
+        myos.pod_read()
+        myos.get_pvc()
+
+  pvcscale.yml: |
+    provider: local
+    ClusterLoader:
+      cleanup: {{ pvcscale_cleanup }}
```
akrzos: `pvcscale_cleanup` seems to be missing.  (For testing purposes I populated the associated vars/pvcscale.yml with the var)

ekuric: that is planned like that. Project needs to be there as long as result.sh is running - as we need to collect pods create time from pvcscale0 project 
I am deleting project at end of test https://github.com/openshift-scale/workloads/blob/2b041f530acb051e00a8ab8bd9a6afea916e4f4f/workloads/pvcscale.yml#L142

akrzos: Ok got it.  Lets change the var into the value directly then (`false`).

Example:
```
pvcscale.yml: |
    provider: local
    ClusterLoader:
      cleanup: false
...
```

workloads/pvcscale.yml
```diff
@@ -0,0 +1,144 @@
...
+      delay: 10
+      until: job_json.stdout | from_json | json_query('status.succeeded==`1` || status.failed==`1`')
+      failed_when: job_json.stdout | from_json | json_query('status.succeeded==`1`') == false
+      when: job_completion_poll_attempts|int > 0
+
+    - name: remove user secret for user useroot - we needed it only for pvc scale test
+      shell: |
+        oc delete secret htpass-secret -n openshift-config
+
+    - name: delete PVC project after test is done
```
akrzos: I am not sure this is a good idea in that an"unspoken" feature is to set the `job_completion_poll_attempts` to 0 so the playbook just kicks off the job and then in CI we can `oc logs -f` the workload container for workload status.  I'd rather see this be a parameter-ized feature "in-band" of the script with an example being like the network tests.  See this Example: https://github.com/openshift-scale/workloads/blob/master/workloads/files/workload-network-script-cm.yml#L104-L123

This permits us to debug the script (Set cleanup to false) and retain the CI ability to tail the logs of the workload container.

workloads/templates/workload-pvcscale-script-cm.yml.j2
ekuric: I see now, fixed

workloads/pvcscale.yml
ekuric: Ok, it can be paramter  to satisfy CI part, I will update

akrzos: Got it thanks, my apologies about the the "unspoken" ci requirements.  The other reason to do it inband of the workload script is such that CI can still run it and it will still cleanup when it is done.

ekuric: for this test, if someone set `cleanup: true` then it is going to clean up project before data are collected, so I had to make it impossible to clean up project with pods/pvcs and delete it at end of test when it is not any more necessary. `cleanup: true` is part of clusterloader and in first step it creates objects and if `cleanup: true` it cleans them at end of `clusterloader` part, timestamp collection is outside of clusterloader section.

akrzos: Got it, so we should allow cleanup, but after the post processing component.  I imagine it would be around [line 55](https://github.com/openshift-scale/workloads/pull/24/files#diff-0039e9acca516519e319ae6e828a4020R55).

ekuric: yes, I will test that scenario and update PR.

workloads/templates/workload-pvcscale-script-cm.yml.j2
```diff
@@ -0,0 +1,367 @@
...
+    # Test Configuration
+    echo "$(date -u) Running PVC scale test"
+    pbench-user-benchmark --config="{{ pvcscale_test_prefix }}-pods-{{ pvcscale_maxpods }}-sc-{{ pvcscale_storageclass }}-create_pods" -- 'VIPERCONFIG=/root/workload/pvcscale.yml openshift-tests run-test "[Feature:Performance][Serial][Slow] Load cluster should load the cluster [Suite:openshift]"'
+
+    echo "$(date -u) Completed PVC scale test"
+    # End Test Configuration
+    echo "Running pvc data collection"
+
+    oc create -f /root/workload/pvcrole.yml
+    oc create -f /root/workload/pvcrolebinding.yml
```
akrzos: These two commands should use replace with --force flag to ensure the script does not break on re-run(When role/rolebinding already exists)
Example:
```
    oc replace --force -f /root/workload/pvcrole.yml
    oc replace --force -f /root/workload/pvcrolebinding.yml
```

ekuric: changed to `replace` instead of `create`

ekuric:
documentation #46 @akrzos

workloads/templates/workload-pvcscale-script-cm.yml.j2
akrzos: I believe Jiri was referring to the spacing.  It is a nitpick but makes for cleaner and more readable code.

jmencak: yes, consistency, please, spacing

ekuric: fixed, :), thanks @jmencak

ekuric:
@akrzos ptal/test , ty.

akrzos:
The workload runs both with and without tooling at this point, I believe it is a good time to merge and mark any outstanding items as an issue to be addressed later and/or by other team members.  Thanks Elko for the effort here!  I will detail some of the remaining issues/items in a separate issue to track that work going forward.

PR #27, opened 2019-06-17
Fix the node selector variable default as a boolean

akrzos:
WORKLOAD_JOB_NODE_SELECTOR is a boolean variables thus it's default should be a boolean instead of an empty string


chaitanyaenr (approved):
LGTM

PR #29, opened 2019-06-18
Add a docs directory and improve existing docs

akrzos:
Here is some more detailed work to address #16 @chaitanyaenr @ekuric @jmencak @sjug PTAL


docs/README.md
```diff
@@ -0,0 +1,22 @@
+# Table of workloads
+
+| Workload/tooling                                   | Short Description                         | Minimum Requirements                  |
+|:-------------------------------------------------- |:----------------------------------------- | ------------------------------------- |
+| [Tooling](tooling.md)                              | Setup pbench metric tools                 | Cluster-admin,Privileged Containers   |
```
jmencak: Suggest keep things consistent: `admin,Privileged` -> `admin, Privileged`

docs/README.md
```diff
@@ -0,0 +1,22 @@
...
+| [Test](test.md)                                    | Test/Run your workload from ssh Container | Cluster-admin,Privileged Containers   |
```
jmencak: same as above

jmencak (commented):
This is great, thanks for doing this!  The only suggestion I have is to keep the section names "All environment variables" vs. "RHCOS workload variables" consistent and use only one.  Even better, though, do we need these sections at all?  I'd suggest having only one section and explain those variables in a table, like most of the OpenShift docs used to do (e.g. [Router Environment Variables](https://docs.openshift.com/container-platform/3.11/architecture/networking/routes.html#env-variables)). Repeating the variable names in two sections is just duplication and extra work for folks searching for meaning of the variables.  Instead, I'd suggest having a "suggested configuration(s)" instead for every workload.  For now, this can be empty and folks responsible for "their" workload should add this.

akrzos:
Thanks Jiri, Let me see if I can have a multi-line table with markdown.  I like the idea of having an explicit default column and simplify to a single table.

akrzos:
@jmencak PTAL again.  I left the http section not updated, perhaps we can merge this in and you can provide a P/R to update to the latest vars as well as your preferred smoke test variables as well? (And any other adjustments or further description you would like to provide)

@chaitanyaenr I left the space on the "front page" README for how scale-ci-pipeline will generate this as a jenkins jobs.  Once you start adding those jobs, can you provide a P/R to populate that space?

Lastly please review for catching any mistakes and lets get it merged!

chaitanyaenr:
@akrzos sorry for the delay. The docs section looks good to me, will create a PR to update the readme about scale-ci-pipeline. Thanks.

Issue #2, opened 2019-05-29
Verify pbench daemonset mounts are in use


akrzos:
There are several mounts in the pbench daemonsets that it is unclear if those mounts are in use or are left over cruft. 

Example: https://github.com/openshift-scale/workloads/blob/master/workloads/templates/pbench-master-ds.yml.j2#L44-L49

akrzos:
@chaitanyaenr could you advise which mounts are still in use and which are no longer (if any so we can clean up the mounts)?

chaitanyaenr:
@akrzos on it.

akrzos:
@chaitanyaenr any updates here?

akrzos:
Fixed with #31

PR #31, opened 2019-06-20
Cleanup the unwanted daemonsets mounts

chaitanyaenr:
We used to mount proc and certs from te host onto the agent containers
for some of the tools like prometheus metrics, we no longer need them
as we are using scale-ci-promdump and scraper for collection and analysis.


chaitanyaenr:
Fixes https://github.com/openshift-scale/workloads/issues/2

akrzos (approved):
LGTM Thanks @chaitanyaenr

PR #38, opened 2019-06-21
No need to specify the image twice anymore

akrzos:
We only need to specify the image in the pod yaml for these two tests.


chaitanyaenr (approved):
LGTM

PR #41, opened 2019-06-22
Set KUBECONFIG environment variable

chaitanyaenr:
This is needed for the client to talk to apiserver when the kubeconfig
is not at the default path i.e $HOME/.kube/config.

### DIFF
--- a/workloads/baseline.yml
+++ b/workloads/baseline.yml
@@ -11,6 +11,8 @@
     - vars/baseline.yml
   vars:
     workload_job: "baseline"
+  environment:
+    KUBECONFIG: "{{ kubeconfig_file }}"
   tasks:
     - name: Create scale-ci-tooling directory
       file:
--- a/workloads/cluster-limits-deployments-per-ns.yml
+++ b/workloads/cluster-limits-deployments-per-ns.yml
@@ -11,6 +11,8 @@
     - vars/deployments-per-ns.yml
   vars:
     workload_job: "deployments-per-ns"
+  environment:
+    KUBECONFIG: "{{ kubeconfig_file }}"
   tasks:
     - name: Create scale-ci-tooling directory
       file:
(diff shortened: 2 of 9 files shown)


workloads/baseline.yml
```diff
@@ -11,6 +11,8 @@
     - vars/baseline.yml
   vars:
     workload_job: "baseline"
+  environment:
+    KUBECONFIG: "{{ kubeconfig_file }}"
```
akrzos, line 15: Should `kubeconfig_file` be set in every vars/$WORKLOAD.yml file?

chaitanyaenr: @akrzos We set the kubeconfig_file in every workload vars file but if it's not set to ~/.kube/config the oc commands executed outside the scale-ci-workload pod won't work as it either looks at $HOME/.kube/config or KUBECONFIG env var or we need to use $oc --config <path-to-kubeconfig>. Setting the KUBECONFIG env var to the path mentioned in the workloads vars file i.e kubeconfig_file will be useful.

akrzos: Oops, I managed to miss that.  Curious though, is there a specific CI need for this at the moment or is this simply to help those who don't have kubeconfig in a standard place?  I can see this helping if you deploy multiple clusters from your local machine and want to run multiple workload repos for each cluster (Thus your kubeconfig file would be in different places for each cluster).  Is that the intended fix?

chaitanyaenr: That's one of the use case, we also need this for integrating Scale-CI into starter framework as the kubeconfig on starter clusters is not in the home directory.

akrzos: Got it. OK so we need to change the default as well since this will "override" where the kubeconfig "is".  Since tilde expansion will not occur inside the `environment` section for Ansible, we need to set the default to what is the most usual case or most usual un-automated case imo.  An example would be change it to `/root/.kube/config` since the `ORCHESTRATION_USER` default is `root` already and that is where the tilde would expand it to that case.  Thoughts?

akrzos:
@chaitanyaenr is this one still relevant?

chaitanyaenr:
@akrzos this can be closed. We can open it again in case there's a need.

PR #46, opened 2019-06-28
pvcscale.md initial document

ekuric:
pvcscale.md initial document @akrzos @jmencak @chaitanyaenr PTAL

### DIFF
--- a/docs/pvcscale.md
+++ b/docs/pvcscale.md
@@ -1,0 +1,232 @@
+# PVCScale Workload
+
+The PVCScale workload playbook is `workloads/pvcscale.yml` and will run the pvcscale workload on your cluster.
+
+PVCScale workload stresses storage backend and determines how fast Pods are started depending on storage backend. 
+Each Pod requires a PVC to be allocated and attached.
+
+PVCScale test creates a desired number of Pods and attaches one PVC to each Pod, it supports 
+only dynamic PVC allocation using storage class and in order for this test to 
+work it is a requirement to have a working storage class which can be used to allocate a PVC.
+
+At end of test, the time necessary for Pods to reach `READY` state with PVC
+in `Bound` status is measured, and based on this time we can compare start time
+for same number of Pods/PVCS for cases when only difference is storage class
+used to allocate PVCS. 
+
+The situation when Pod is in `READY` state and  with PVC in `Bound` status is considered as state where PVC is
+ready for consumption by application Pod.
+
+Different storage backends behaves differently when it comes to PVC scalability and it is necessary to have 
+repeatable results which can answer the question what storage backend provides optimal PVCs scale and density results for
+scenario where many Pods/PVCs are necessary to start in short period of time.
+
+Running from CLI:
+
+```sh
+$ cp workloads/inventory.example inventory
+$ # Add orchestration host to inventory
+$ # Edit vars in workloads/vars/pvcscale.yml or define Environment vars (See below)
+$ time ansible-playbook -vv -i inventory workloads/pvcscale.yml
+```
+
+## Environment variables
+
+### PUBLIC_KEY
+Default: `~/.ssh/id_rsa.pub`
+
+Public ssh key file for Ansible.
+
+### PRIVATE_KEY
+Default: `~/.ssh/id_rsa`
+
+Private ssh key file for Ansible.
+
+### ORCHESTRATION_USER
+Default: `root`
+
+User for Ansible to log in as. Must authenticate with PUBLIC_KEY/PRIVATE_KEY.
+
+### WORKLOAD_IMAGE
+Default: `quay.io/openshift-scale/scale-ci-workload`
+
+Container image that runs the workload script.
+
+### WORKLOAD_JOB_NODE_SELECTOR
+Default: `''`
+
+Enables/disables the node selector that places the workload job on the `workload` node.
+
+### WORKLOAD_JOB_TAINT
+Default: `true`
+
+Enables/disables the toleration on the workload job to permit the `workload` taint.
+
+### WORKLOAD_JOB_PRIVILEGED
+Default: `true`
+
+Enables/disables running the workload Pod as privileged.
+
+### KUBECONFIG_FILE
+Default: `~/.kube/config`
+
+Location of kubeconfig on orchestration host. PVCScale tests supports only
+default location for `kubeconfig` file.
+
+### PBENCH_SSH_PRIVATE_KEY_FILE
+Default: `~/.ssh/id_rsa`
+
+Location of ssh private key to authenticate to the pbench results server.
+
+### PBENCH_SSH_PUBLIC_KEY_FILE
+Default: `~/.ssh/id_rsa.pub`
+
+Location of the ssh public key to authenticate to the pbench results server.
+
+### ENABLE_PBENCH_AGENTS
+Default: `false`
+
+Enables/disables the collection of pbench data on the pbench agent Pods. These Pods are deployed by the workloads/tooling.yml playbook
+it needs to be executed independently of PVCScale test.
+
+### PBENCH_SERVER
+Default: There is no public default.
+
+DNS address of the pbench results server.
+
+### SCALE_CI_RESULTS_TOKEN
+Default: There is no public default.
+
+Future use for pbench and prometheus scraper to place results into git repo that holds results data.
+
+### JOB_COMPLETION_POLL_ATTEMPTS
+Default: `360`
+
+Number of retries for Ansible to poll if the workload job has completed. Poll attempts delay 10s between polls with some additional time taken for each polling action depending on the orchestration host setup.
+
+### PVCSCALE_TEST_PREFIX
+Default: `pvcscale`
+
+Test to prefix the pbench results.
+
+### PVCSCALE_CLEANUP
+Default: `true`
+
+Enables/disables cleanup of this workload on completion. Cleanup in
+PVCScale terminology means deletion of `pvcscale0` project - this operation can be
+resource intensive in case there are many Pods/PVCs  in `pvcscale0` project which are created
+during test.
+
+### PVCSCALE_BASENAME
+Default: `pvcscale`
+
+Basename used by cluster loader for the project(s) it creates.
+
+### PVCSCALE_MAXPODS
+Default: `10`
+
+Maximum number of Pods that will be created by the pvcscale workload. 
+Typically much higher values are used than the default for large scale tests.
+PVCScale does not have implemented mechanism to calculate maximum number of Pods which all allowed to be started on 
+particular cluster, so it is necessary to adapt `PVCSCALE_MAXPODS` to cluster capacity and
+execute it accordingly. 
+Choosing `PVCSCALE_MAXPODS` to be high can lead that many Pods/PVCs are created what can affect cluster limits and peformance.
+
+### PVCSCALE_POD_IMAGE
+Default: `quay.io/openshift-scale/scale-ci-workload`
+
+Pod container image to use for Pods that are pvscale Pods. This image will be used to
+start Pod and bound PVC to it.
+
+### PVCSCALE_STEPSIZE
+Default: `1`
+
+Number of Pods for cluster loader will create before waiting for Pods to become running.
+PVCScale test creates one Pod/PVC at time.
+
+### PVCSCALE_PAUSE
+Default: `0`
+
+Period of time (in seconds) for cluster loader to pause after creating Pods and 
+waiting for them to be running.PVCScale does not wait on Pods to be `READY` state
+before it creates new ones.
+
+### PVCSCALE_STORAGE_SIZE
+Default: `1Gi`
+
+This is default PVC size. If specified as `1Gi` then all Pods will get `1Gi` PVC
+mounted to `/mnt/pvcmount` inside Pod.
+
+### PVCSCALE_STORAGECLASS
+Default: `none`
+
+The storageclass which will be used to allocated PVCs. It has to be present on cluster and
+must be functional for test to work. Deafault is `none` and must be changed prior running test 
+and adapted to name of functional storage class name.
+
+
+### ACCESS_MODES
+Default: `ReadWriteOnce`
+
+`ACCESS_MODES` is access_mode for PVC inside Pod. This field does not
+affect the test itself and might be removed in future. 
+
+
+### Smoke test variables
+
+```
+ORCHESTRATION_HOST="localhost"
+ORCHESTRATION_USER=root
+WORKLOAD_JOB_PRIVILEGED=false
+KUBECONFIG_FILE='~/.kube/config'
+# adapt PUBLIC_KEY and PRIVATE_KEY
+PBENCH_SSH_PRIVATE_KEY_FILE='~/.ssh/id_rsa'
+PBENCH_SSH_PUBLIC_KEY_FILE='~/.ssh/id_rsa.pub'
+ENABLE_PBENCH_AGENTS=false
+PBENCH_SERVER=pbench.dev.openshift.com
+PVCSCALE_TEST_PREFIX=pvcscale
+PVCSCALE_CLEANUP=false
+PVCSCALE_BASENAME=pvcscale
+PVCSCALE_MAXPODS=10
+PVCSCALE_POD_IMAGE="quay.io/openshift-scale/scale-ci-workload"
+PVCSCALE_STEPSIZE=1
+PVCSCALE_PAUSE=0
+PVCSCALE_STORAGE_SIZE="1Gi"
+PVCSCALE_STORAGECLASS="storgage_class_name"
+ACCESS_MODES="ReadWriteOnce"
+# adapt PUBLIC_KEY and PRIVATE_KEY to your needs 
+PUBLIC_KEY=~/.ssh/id_rsa.pub
+PRIVATE_KEY=~/.ssh/id_rsa
+echo "[orchestration]" > inventory
+```
+After exporting above variables, execute test with 
+```
+time ansible-playbook -v -i inventory workloads/workloads/pvcscale.yml
+```
+### Test results 
+
+PVCScale test creates first the number of Pods specified in `PVCSCALE_MAXPods`
+variable. Once all Pods are in `READY` state with PVC in `Bound` status,  PVCScale test 
+collects data for all Pods when they are started, and from this information we can then see
+how long it takes to start Pods from Pod1..PodN.
+
+The output will be like one below 
+```
+Pod-Ready
+07:56:29
+07:56:30
+07:56:31
+07:56:32
+07:56:35
+07:56:36
+07:56:39
+07:56:39
+07:56:44
+07:56:46
+``` 
+In this example are visible times when Pods were in `READY` state, where the
+first Pod is started at `07:56:29` and last Pod at `07:56:46`. The time in this output
+is day time when test was executed, and this will be changed in future versions
+of this test to be independent of time when test was ran. In this case the
+difference between last pod start time and first pod start time is total time
+necessary for Pods to be in `READY` state with PVC in `Bound` status.
--- a/docs/pvcscale.md
+++ b/docs/pvcscale.md
@@ -2,24 +2,15 @@
 
 The PVCScale workload playbook is `workloads/pvcscale.yml` and will run the pvcscale workload on your cluster.
 
-PVCScale workload stresses storage backend and determines how fast Pods are started depending on storage backend. 
-Each Pod requires a PVC to be allocated and attached.
+PVCScale workload stresses storage backend and determines how fast Pods are started depending on storage backend. Each Pod requires a PVC to be allocated and attached.
 
-PVCScale test creates a desired number of Pods and attaches one PVC to each Pod, it supports 
-only dynamic PVC allocation using storage class and in order for this test to 
-work it is a requirement to have a working storage class which can be used to allocate a PVC.
+PVCScale test creates a desired number of Pods and attaches one PVC to each Pod, it supports only dynamic PVC allocation using storage class and in order for this test to work it is a requirement to have a working storage class which can be used to allocate a PVC.
 
-At end of test, the time necessary for Pods to reach `READY` state with PVC
-in `Bound` status is measured, and based on this time we can compare start time
-for same number of Pods/PVCS for cases when only difference is storage class
-used to allocate PVCS. 
+At end of test, the time necessary for Pods to reach `READY` state with PVC in `Bound` status is measured, and based on this time we can compare start time for same number of Pods/PVCS for cases when only difference is storage class used to allocate PVCS.
 
-The situation when Pod is in `READY` state and  with PVC in `Bound` status is considered as state where PVC is
-ready for consumption by application Pod.
+When the Pod is in `READY` state and the PVC in `Bound` status it is considered that the PVC is ready for consumption by application Pod.
 
-Different storage backends behaves differently when it comes to PVC scalability and it is necessary to have 
-repeatable results which can answer the question what storage backend provides optimal PVCs scale and density results for
-scenario where many Pods/PVCs are necessary to start in short period of time.
+Different storage backends behaves differently when it comes to PVC scalability and it is necessary to have repeatable results which can answer the question what storage backend provides optimal PVCs scale and density results for scenario where many Pods/PVCs are necessary to start in short period of time.
 
 Running from CLI:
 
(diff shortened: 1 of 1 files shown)


docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
+# PVCScale Workload
+
+The pvcscale workload playbook is `workloads/pvcscale.yml` and will run the pvcscale workload on your cluster.
+
+PVCScale workload is designed to stress storage backend from the scale point of
+view. It will create desired number of pods and attach an PVC to each pod.
```
jmencak: Perhaps: "It creates a desired number of pods and attaches a PVC to each pod."?

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+During this test we measure time necessary for pods to be in `Ready/Ready` state
```
jmencak: What is `Ready/Ready` state?

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+and for all PVC to be in `Bound` state.
+
+For PVCScale test to work it is necessary that cluster is operational and there
+is working storage class which will be used to carve PVCs from. If storageclass
```
jmencak: there is *a* working storage class?
If *no* storage class?

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+is not present then this test will fail and will not work.
```
jmencak: will fail/will not work (one of them should do).

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+Number of retries for Ansible to poll if the workload job has completed. Poll attempts delay 10s between polls with some additional time taken for each polling action depending on the orchestration host setup.
+
+### PVCSCALE_TEST_PREFIX
+Default: `pvcscale`
+Test to prefix the pbench results.
+
+### PVCSCALE_CLEANUP
+Default: `true`
+Enables/disables cleanup of this workload on completion. Cleanup in
+PVCScale terminology means delete `pvcscale0` projects - this operation can be
```
jmencak: "means deletion of"?
projects &mdash; or &ndash; here depending on your preference?

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+
+### PVCSCALE_BASENAME
+Default: `pvcscale`
+Basename used by cluster loader for the project(s) it creates.
+
+### PVCSCALE_MAXPODS
+Default: `10`
+Maximum number of pods that will be created by the pvcscale workload. 
+Typically much higher values are used than the default for large scale tests.
+PVCScale does not mechanism to calculate maximum number of pods which it can
```
jmencak: Does not have a mechanism?  Or just does not calculate?

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+start, so it is necessary to adapt `PVCSCALE_MAXPODS` to cluster capacity and
```
jmencak: s/adapt/adjust/ ?

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+execute it accordingly. Choosing `PVCSCALE_MAXPODS` to be high can lead that
```
jmencak: Choosing `PVCSCALE_MAXPODS` too high can result in too many pods/PVCs created, which can ... ?  Also consistency PVCs or pvcs (across this document).

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+many pods/PVCs are created what can affect cluster limits and peformance.
+
+### PVCSCALE_POD_IMAGE
+Default: `quay.io/openshift-scale/scale-ci-workload`
+Pod image to use for pods that are pvscale pods. This image will be used to
```
jmencak: Pod image sounds a bit strange to me.  A container image?

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+start pod and bound PVC to it.
+
+### PVCSCALE_STEPSIZE
+Default: `1`
+Number of pods for cluster loader will create before waiting for pods to become running.
```
jmencak: The number of pods cluster loader creates before... ?

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+PVCScale test creates one pod/pvc at time.
+
+### PVCSCALE_PAUSE
+Default: `0`
+Period of time (in seconds) for cluster loader to pause after creating pods and 
+waiting for them to be running.PVCScale does not wait on pods to be `Ready`
```
jmencak: s/.PVCScale/. PVCScale/

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+### PVCSCALE_STEPSIZE
+Default: `1`
+Number of pods for cluster loader will create before waiting for pods to become running.
+PVCScale test creates one pod/pvc at time.
+
+### PVCSCALE_PAUSE
+Default: `0`
+Period of time (in seconds) for cluster loader to pause after creating pods and 
+waiting for them to be running.PVCScale does not wait on pods to be `Ready` 
+before it creates new ones - it creates pods in fastest possible manner.
```
jmencak: -- or ---
in *the* fastest?

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+### PVCSCALE_PAUSE
+Default: `0`
+Period of time (in seconds) for cluster loader to pause after creating pods and 
+waiting for them to be running.PVCScale does not wait on pods to be `Ready` 
+before it creates new ones - it creates pods in fastest possible manner. 
+
+### PVCSCALE_STORAGE_SIZE
+Default: `1Gi`
+This is default PVC size. If specified as `1Gi` then all pods will get `1Gi`
+device mounted inide pods.
```
jmencak: inside

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+
+### PVCSCALE_STORAGECLASS
+Default: `none`
+It is important to have working storage class prior starting PVCScale test.
```
jmencak: to have *a* working ...?
starting *the* PVCScale... ?

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+`PVCSCALE_STORAGECLASS` parameter is name of storage class which one plan to
```
jmencak: is *the* name of *the* storage class to be used ?
and one plan*s* to

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+use, and any storage class supporting dynamic storage provisioning will work
```
jmencak: unfinished sentence?

This sentence needs a bit more work, not sure I understand it.

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+
+### ACCESS_MODES
+Default: `ReadWriteOnce`
+`ACCESS_MODES` is access_mode for PVC inside container. This field does not
```
jmencak: inside which container?

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+affect test itself and might be removed in future.
```
jmencak: affect *the* test

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+PVCSCALE_CLEANUP=true
+PVCSCALE_BASENAME=pvcscale
+PVCSCALE_MAXPODS=10
+PVCSCALE_POD_IMAGE="quay.io/openshift-scale/scale-ci-workload"
+PVCSCALE_STEPSIZE=1
+PVCSCALE_PAUSE=0
+```
+
+### Test results 
+PVCScale test will first create number of pods specified in `PVCSCALE_MAXPODS`.
```
jmencak: create *a* number of pods

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+Once all pods are in `Ready/Ready` state with bound PVC, PVCScale test will
```
jmencak: Again, not sure what `Ready/Ready` is.

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+07:56:32
+07:56:35
+07:56:36
+07:56:39
+07:56:39
+07:56:44
+07:56:46
+``` 
+In this example are visible times when pods were in "Ready/Ready" state, where
+first pods started at `07:56:29` and last at `07:56:46`. The time is relative
```
jmencak: the first pods.
Also, I'm not sure I understand the "The time is relative...".

docs/pvcscale.md
```diff
@@ -0,0 +1,179 @@
...
+07:56:31
+07:56:32
+07:56:35
+07:56:36
+07:56:39
+07:56:39
+07:56:44
+07:56:46
+``` 
+In this example are visible times when pods were in "Ready/Ready" state, where
```
jmencak: `Ready/Ready`

jmencak (commented):
Not a native English speaker, so please take my comments only as suggestions.  On the whole looking good, I only didn't understand a few sentences.

akrzos (changes requested):
Please add an entry into the docs/README.md table for pvcscale - https://github.com/openshift-scale/workloads/blob/master/docs/README.md#table-of-workloads

docs/pvcscale.md
ekuric: can be like this or something else :), will rework this part anyway.

docs/pvcscale.md
ekuric: READY (1/1) - passed successfully  all liveness and readiness probes and started , will term READY instead of Ready/Ready

ekuric:
@akrzos @jmencak ptal, this goes with #49

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+At end of test, the time necessary for PODS to reach `READY` state with PVC
+in `Bound` status is measured, and based on this time we can compare start time
+for same number of PODS/PVCS for cases when only difference is storage class
+used to allocate PVCS. 
+
+The situation when POD is in `READY` state and  with PVC in `Bound` status is considered as state where PVC is
+ready for consumption by application POD.
+
+
+Different storage backens bahaves diferently when it comes to PVC scalability and it necessary to have
```
jmencak: s/backens/backends/
s/bahaves/behaves/
s/diferently/differently/
s/it necessary/it is necessary/

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+07:56:36
+07:56:39
+07:56:39
+07:56:44
+07:56:46
+``` 
+In this example are visible times when PODS were in `READY` state, where the
+first pod is started at `07:56:29` and last one at `07:56:46`.The time in this output
+is day time when test was executed, and this will be changed in future versions
+of this test to be indenpendent of time when test was ran.
```
jmencak: s/indenpendent/independent/

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
+# PVCScale Workload
+
+The PVCScale workload playbook is `workloads/pvcscale.yml` and will run the pvcscale workload on your cluster.
+
+PVCScale workload is designed to stress storage backend and provide information
+how fast PODS are stared depending on storage backend they require PVCs to be
```
jmencak: I'd suggest s/POD/Pod/ in this document.  See upstream k8s docs.

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+allocated from and attached to them.
+
+PVCScale test creates a desired number of PODS and attaches one PVC to each POD, it supports
```
jmencak: s/POD, it/Pod. It/

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+
+The PVCScale workload playbook is `workloads/pvcscale.yml` and will run the pvcscale workload on your cluster.
+
+PVCScale workload is designed to stress storage backend and provide information
+how fast PODS are stared depending on storage backend they require PVCs to be
+allocated from and attached to them.
+
+PVCScale test creates a desired number of PODS and attaches one PVC to each POD, it supports 
+only dynamic PVC allocation using storage class and in order for this test to 
+work it is requirement to have working storage class which can be used to allocate PVC.
```
jmencak: s/it is requirement/it is a requirement/
to have *a* working storage class
used to allocate *a* PVC (or used to allocate PVCs?)

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+
+At end of test, the time necessary for PODS to reach `READY` state with PVC
+in `Bound` status is measured, and based on this time we can compare start time
+for same number of PODS/PVCS for cases when only difference is storage class
+used to allocate PVCS. 
+
+The situation when POD is in `READY` state and  with PVC in `Bound` status is considered as state where PVC is
```
jmencak: *a* Pod is in ... ?
s/[ ]{2,}/ /

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+ready for consumption by application POD.
```
jmencak: *an* application Pod ?

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+
+
+Different storage backens bahaves diferently when it comes to PVC scalability and it necessary to have 
+repeatable results which can answer the question what storage backend provides optimal PVCS scale and density results for
+scenario where many PODS/PVC are necessary to start in short period of time.
```
jmencak: in *a* short ?

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+Number of retries for Ansible to poll if the workload job has completed. Poll attempts delay 10s between polls with some additional time taken for each polling action depending on the orchestration host setup.
+
+### PVCSCALE_TEST_PREFIX
+Default: `pvcscale`
+Test to prefix the pbench results.
+
+### PVCSCALE_CLEANUP
+Default: `true`
+Enables/disables cleanup of this workload on completion. Cleanup in
+PVCScale terminology means deletion of `pvcscale0` project - this operation can be
```
jmencak: m/ndash?

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+resource intensive in case there are many PODS/PVCS  in `pvcscale0` project which are created
```
jmencak: s/[ ]{2,}/ /

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+during test.
```
jmencak: *a/the* test?

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+
+### PVCSCALE_BASENAME
+Default: `pvcscale`
+Basename used by cluster loader for the project(s) it creates.
+
+### PVCSCALE_MAXPODS
+Default: `1`
+Maximum number of pods that will be created by the pvcscale workload. 
+Typically much higher values are used than the default for large scale tests.
+PVCScale does not have implemented mechanism to calculate maximum number of pods which all allowed to be started on
```
jmencak: s/all/are/ ?

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+### PVCSCALE_BASENAME
+Default: `pvcscale`
+Basename used by cluster loader for the project(s) it creates.
+
+### PVCSCALE_MAXPODS
+Default: `1`
+Maximum number of pods that will be created by the pvcscale workload. 
+Typically much higher values are used than the default for large scale tests.
+PVCScale does not have implemented mechanism to calculate maximum number of pods which all allowed to be started on 
+particular cluster, so it is necessary to adapt `PVCSCALE_MAXPODS` to cluster capacity and
```
jmencak: *a* particular
adapt/adjust?

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+execute it accordingly. 
+Choosing `PVCSCALE_MAXPODS` to be high can lead that many PODS/PVCs are created what can affect cluster limits and peformance.
```
jmencak: Setting `PVCSCALE_MAXPODS` too high can result in too many Pods/PVCs created, which can affect performance on an under-sized cluster?

docs/pvcscale.md
jmencak: Container image to use for pvscale pods?

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+
+### PVCSCALE_POD_IMAGE
+Default: `quay.io/openshift-scale/scale-ci-workload`
+Pod image to use for pods that are pvscale pods. This image will be used to
+start POD and bound PVC to it.
```
jmencak: start pvscale Pods and bound PVC to them
or
start a pvscale Pod and bound a PVC to it
?

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+
+### PVCSCALE_STEPSIZE
+Default: `1`
+Number of PODS for cluster loader will create before waiting for pods to become running.
```
jmencak: Number of Pods for [Cluster Loader](https://docs.openshift.com/container-platform/4.1/scalability_and_performance/using-cluster-loader.html) to create ...

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+PVCScale test creates one pod/pvc at time.
```
jmencak: one Pod (consistency, now PODs are everywhere, suggest Pods).
s/at time/at a time/
?

docs/pvcscale.md
akrzos: s/stared/started

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+
+### PVCSCALE_PAUSE
+Default: `0`
+Period of time (in seconds) for cluster loader to pause after creating pods and 
+waiting for them to be running.PVCScale does not wait on pods to be `Ready` 
+before it creates new ones - it creates pods in fastest possible manner. 
+
+### PVCSCALE_STORAGE_SIZE
+Default: `1Gi`
+This is default PVC size. If specified as `1Gi` then all pods will get `1Gi`
```
jmencak: *the* default?
then all pods get ...  I believe that we Europeans tend to overuse "*will*". Alex can probably comment.

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+### PVCSCALE_STORAGECLASS
+Default: `none`
+It is important to have working storage class prior starting PVCScale test.
+`PVCSCALE_STORAGECLASS` parameter is name of storage class which one plan to
+use, and any storage class supporting dynamic storage provisioning will work
+
+### ACCESS_MODES
+Default: `ReadWriteOnce`
+`ACCESS_MODES` is access_mode for PVC inside container. This field does not
+affect the test itself and might be removed in future.
```
jmencak: in *the* future

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+PVCSCALE_TEST_PREFIX=pvcscale_smoke
+PVCSCALE_CLEANUP=true
+PVCSCALE_BASENAME=pvcscale
+PVCSCALE_MAXPODS=10
+PVCSCALE_POD_IMAGE="quay.io/openshift-scale/scale-ci-workload"
+PVCSCALE_STEPSIZE=1
+PVCSCALE_PAUSE=0
+```
+
+An longer example of variables can be found
```
jmencak: *A* longer

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+[here](https://gist.githubusercontent.com/ekuric/6038cb7437d58679de00d31b0365ce4c/raw/d7570f0a4499c6f9d9a395ecdd969bf3bc8b59b3/gistfile1.txt)
```
jmencak: Not sure we should use pastes like this, @akrzos ?  Not a fan myself.

docs/pvcscale.md
akrzos: I think this would be clearer:

PVCScale workload stresses storage backend and determines how fast pods are started depending on storage backend. Each pod requires a PVC to be allocated and attached.

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Test results 
+PVCScale test creates first the number of PODS specified in `PVCSCALE_MAXPODS`
```
jmencak: s/creates first/first creates/ ? (Not sure here, but sounds more *right*).

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+variable. Once all PODS are in `READY` state with PVC in `Bound` status,  PVCScale test 
+collects data for all PODS when they are started, and from this information we can then see
+how long it takes to start PODS from pod1..podN.
+
+The output will be like one below
```
jmencak: See sample output from XYZ below

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+07:56:32
+07:56:35
+07:56:36
+07:56:39
+07:56:39
+07:56:44
+07:56:46
+``` 
+In this example are visible times when PODS were in `READY` state, where the
+first pod is started at `07:56:29` and last one at `07:56:46`.The time in this output
```
jmencak: s/.The/. The/

docs/pvcscale.md
akrzos: I think POD can just be referred to as pod and pvc as pvc or we should keep it consistent throughout all docs if we refer to pod as POD etc... (I'll have to check other docs if we do otherwise.)

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+07:56:31
+07:56:32
+07:56:35
+07:56:36
+07:56:39
+07:56:39
+07:56:44
+07:56:46
+``` 
+In this example are visible times when PODS were in `READY` state, where the
```
jmencak: Would "The output shows the first Pod was in `READY` state at `07:56:29`, the last one at `07:56:46`." be better?

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+first pod is started at `07:56:29` and last one at `07:56:46`.The time in this output
+is day time when test was executed, and this will be changed in future versions
```
jmencak: is a local time.
That's it.  I would not comment on changing anything in future versions here, even though you will.  If you insist, you have a Future work section.

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+of this test to be indenpendent of time when test was ran. 
+
+### Future work 
+
+- Improve test results to show start times starting at `00:00:00` or similar
+  approach and change current representation showing time of day when test was
+executed.
+- rewrite [check_pvpods.py](https://raw.githubusercontent.com/ekuric/openshift/master/cns/check_pods/check_pvpods.py) in golang
```
jmencak: This is a link to an external repo.  Can you link a openshift-scale repo that has this file?

jmencak (changes requested):
Good overall, nits + I'd remove links to external repos though as we want to keep everything in openshift-scale now.

docs/pvcscale.md
ekuric: yup, fixed,

docs/pvcscale.md
ekuric: fixed

docs/pvcscale.md
ekuric: very good suggestion, I forgot about this

docs/pvcscale.md
```diff
@@ -0,0 +1,197 @@
...
+$ cp workloads/inventory.example inventory
+$ # Add orchestration host to inventory
+$ # Edit vars in workloads/vars/pvcscale.yml or define Environment vars (See below)
+$ time ansible-playbook -vv -i inventory workloads/pvcscale.yml
+```
+
+## Environment variables
+
+### PUBLIC_KEY
+Default: `~/.ssh/id_rsa.pub`
```
akrzos: Each of these (Each env var default) needs 2 spaces after the "default" this will cause a break to be between the Default and the text below the default.  Currently markdown consumes the newline without these two spaces, take a look at the other readme files to see an example.

ekuric: @akrzos fixed,

docs/pvcscale.md
```diff
@@ -0,0 +1,221 @@
...
+
+User for Ansible to log in as. Must authenticate with PUBLIC_KEY/PRIVATE_KEY.
+
+### WORKLOAD_IMAGE
+Default: `quay.io/openshift-scale/scale-ci-workload`
+
+Container image that runs the workload script.
+
+### WORKLOAD_JOB_NODE_SELECTOR
+Default: `true`
```
akrzos: The default in the vars file is actually set to `''` right now, we should fix the default in the vars file to match here.

docs/pvcscale.md
```diff
@@ -0,0 +1,221 @@
...
+Location of the ssh public key to authenticate to the pbench results server.
+
+### ENABLE_PBENCH_AGENTS
+Default: `false`
+
+Enables/disables the collection of pbench data on the pbench agent Pods. These Pods are deployed by the workloads/tooling.yml playbook
+it needs to be executed independently of PVCScale test.
+
+### PBENCH_SERVER
+Default: `There is no public default.`
```
akrzos: Since this is no real value I think the backticks surrounding the quote "There is no public default." should be removed.

docs/pvcscale.md
```diff
@@ -0,0 +1,221 @@
...
+
+DNS address of the pbench results server.
+
+### SCALE_CI_RESULTS_TOKEN
+Default: `There is no public default.`
```
akrzos: Same as above ^, Since this is no real value I think the backticks surrounding the quote "There is no public default." should be removed.

docs/pvcscale.md
```diff
@@ -0,0 +1,221 @@
...
+resource intensive in case there are many Pods/PVCs  in `pvcscale0` project which are created
+during test.
+
+### PVCSCALE_BASENAME
+Default: `pvcscale`
+
+Basename used by cluster loader for the project(s) it creates.
+
+### PVCSCALE_MAXPODS
+Default: `1`
```
akrzos: The actual default is `10` in the vars file.  I am not sure which value is more preferable.

docs/pvcscale.md
```diff
@@ -0,0 +1,221 @@
...
+before it creates new ones.
+
+### PVCSCALE_STORAGE_SIZE
+Default: `1Gi`
+
+This is default PVC size. If specified as `1Gi` then all Pods will get `1Gi` PVC
+mounted to `/mnt/pvcmount` inside Pod.
+
+### PVCSCALE_STORAGECLASS
+Default: `none`
```
akrzos: The default in the vars file is currently `csi-rdb`, I am not sure which is preferred as the actual default.

docs/pvcscale.md
akrzos: I think we should just shorten the content in the gist, remove anything which shouldn't be exposed and place inline to readme here.

docs/pvcscale.md
```diff
@@ -0,0 +1,221 @@
...
+07:56:46
+``` 
+In this example are visible times when Pods were in `READY` state, where the
+first Pod is started at `07:56:29` and last Pod at `07:56:46`. The time in this output
+is day time when test was executed, and this will be changed in future versions
+of this test to be independent of time when test was ran. In this case the
+difference between last pod start time and first pod start time is total time
+necessary for Pods to be in `READY` state with PVC in `Bound` status.
+
+### Future work
```
akrzos: What about tracking these as an Issue instead of in the readme?

akrzos (changes requested):
I found a few small nitpick issues that we should fix before merging.  Thanks for the content @ekuric

docs/pvcscale.md
ekuric: done

docs/pvcscale.md
ekuric: done

docs/pvcscale.md
ekuric: fixed

docs/pvcscale.md
ekuric: fixed!

docs/pvcscale.md
ekuric: fixed

docs/pvcscale.md
ekuric: ack, fixed.

docs/pvcscale.md
ekuric: I tend to keep some low value just for people to run test and get feeling what is going on with this test, let's keep 10 - which is low value ( 10 Pods/PVC on any cluster is fine )

docs/pvcscale.md
ekuric: ack, #63 and #64

docs/pvcscale.md
ekuric: I am not sure what would be best default here as storage class can be anything, if I leave `csi-rbd` and if not present , test will fail, but that is case for anything if dynamic provisioning is not in place and there is not functional storage class. I expect user to adapt this value prior running test.

akrzos: Got it if we expect the user to change the value as there is no reasonable default lets set this to:

"Default: none"  (Without "`")

And indicate in the description below the "default" choices possible considering the cluster configuration.  Ex: Use `gp2` if your cluster is built on AWS and `csi-rdb` if you have ceph rook installed.

akrzos:
I made a few small adjustments so we can get this merged.  Thanks @ekuric

PR #47, opened 2019-06-29
Add Scaling workload

akrzos:


akrzos:
DNM yet, Thanks

akrzos:
Ok This is ready.  It includes a var for pass/fail criteria (EXPECTED_SCALE_DURATION) . I would like to use this model in conjunction with the scraper and comparison tooling for all future pass/fail test criteria.

akrzos:
@jmencak @chaitanyaenr PTAL

chaitanyaenr (approved):
LGTM, worked fine on a test cluster. We need to make sure is that if we provision a cluster with 3 workers to start we might have to plan for that logic like 3 azs instead of 4 like @akrzos mentioned.

PR #50, opened 2019-07-01
Add Pass/Fail functionality to NodeVertical

akrzos:
Current checks if cluster loader exits correctly


akrzos:
DNM this is a WIP for right now however please examine and provide feedback if you have such.

akrzos:
This is dependent upon this functionality being added to cluster loader - https://github.com/openshift/origin/pull/23310

Otherwise we can have situations where NodeVertical workload will hang on a cluster.  The timeout will prevent those situations.

docs/nodevertical.md
```diff
@@ -35,21 +35,33 @@ Default: `quay.io/openshift-scale/scale-ci-workload`
...
 ### WORKLOAD_JOB_PRIVILEGED
-Default: `true`  
+Default: `false`  
 Enables/disables running the workload pod as privileged.
 
 ### KUBECONFIG_FILE
 Default: `~/.kube/config`  
 Location of kubeconfig on orchestration host.
 
+### PBENCH_INSTRUMENTATION
```
chaitanyaenr, line 53: The docs of other workloads needs to be modified with the new vars ( PBENCH_INSTRUMENTATION and ENABLE_PBENCH_COPY ). Is the plan to create a PR later?

chaitanyaenr:
Looks good, will give it a try on test cluster and we can merge this.

docs/nodevertical.md
akrzos: Eventually yes however those workloads do not implement these vars in the same fashion yet.  I am a bit conflicted on structuring the pbench_* vars as there are a few combinations we will want. (`PBENCH_INSTRUMENTATION`, `ENABLE_PBENCH_AGENTS`, and `ENABLE_PBENCH_COPY`)  For now lets keep this with the nodevertical workload and I will see about extending it to the podvert and mastervert workloads as well.  It is a little bit unclear how we can run with the network tests but I will explore this more once we land this.

The other thought is setting defaults to not require pbench and to be more or less for CI runs.  Thanks for reviewing this.

chaitanyaenr (approved):
Tested it, LGTM.

docs/nodevertical.md
chaitanyaenr: Got it.

PR #52, opened 2019-07-08
Wait until workload pod is Running instead of Job active

akrzos:
Turns out the job can be active but the pod not actually running. Wait until the pod is running allows CI to stream pod logs with little effort.


akrzos:
@chaitanyaenr @jmencak @ekuric PTAL

chaitanyaenr (approved):
LGTM, this will eliminate the need to add a sleep or logic to check the status of the pods when poll attempts until the job is set to 0 to stream the pod logs.

PR #55, opened 2019-07-11
Consolidate the workload env vars into a single jinja2 template

akrzos:
Evenetually any shared vars can be added as a single var outside the if statement


jmencak:
LGTM, unable to test at the moment though.

jmencak:
Tested, worked fine.

PR #56, opened 2019-07-11
Add Pass/Fail Criteria to PodVertical Cluster Limit Test

akrzos:


chaitanyaenr (commented):
Think it would be helpful if we add the functionality to fail the test if it takes maybe 50% less time than expected duration. Maybe it's out of scope for this PR.

akrzos:
> Think it would be helpful if we add the functionality to fail the test if it takes maybe 50% less time than expected duration. Maybe it's out of scope for this PR.

I think you intent here is to detect if say something went wrong but the benchmark still executed.  I would rather focus on having the test tool detect that something went wrong rather than flag off test duration since that would be a coarse indicator imo.

chaitanyaenr (approved):
Tested and works as expected.

PR #59, opened 2019-07-12
initial fio i/o test

ekuric:
@akrzos if you have time, please review/test

### DIFF
--- a/workloads/fio.yml
+++ b/workloads/fio.yml
@@ -1,0 +1,141 @@
+---
+#
+# Runs PVC Scale test on an existing RHCOS cluster
+#
+
+- name: Runs PVC scale on a RHCOS cluster
+  hosts: orchestration
+  gather_facts: true
+  remote_user: "{{orchestration_user}}"
+  vars_files:
+    - vars/fio.yml
+  vars:
+    workload_job: "storage-fio"
+  tasks:
+    - name: Create scale-ci-tooling directory
+      file:
+        path: "{{ansible_user_dir}}/scale-ci-tooling"
+        state: directory
+
+    - name: Copy workload files
+      copy:
+        src: "{{item.src}}"
+        dest: "{{item.dest}}"
+      with_items:
+        - src: scale-ci-tooling-ns.yml
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/scale-ci-tooling-ns.yml"
+          # - src: workload-fio-script-cm.yml
+          # dest: "{{ansible_user_dir}}/scale-ci-tooling/workload-fio-script-cm.yml"
+
+    - name: Slurp kubeconfig file
+      slurp:
+        src: "{{kubeconfig_file}}"
+      register: kubeconfig_file_slurp
+
+    - name: Slurp ssh private key file
+      slurp:
+        src: "{{pbench_ssh_private_key_file}}"
+      register: pbench_ssh_private_key_file_slurp
+
+    - name: Slurp ssh public key file
+      slurp:
+        src: "{{pbench_ssh_public_key_file}}"
+      register: pbench_ssh_public_key_file_slurp
+
+    - name: Template workload templates
+      template:
+        src: "{{item.src}}"
+        dest: "{{item.dest}}"
+      with_items:
+        - src: pbench-cm.yml.j2
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/pbench-cm.yml"
+        - src: pbench-ssh-secret.yml.j2
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/pbench-ssh-secret.yml"
+        - src: kubeconfig-secret.yml.j2
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/kubeconfig-secret.yml"
+        - src: workload-job.yml.j2
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/workload-job.yml"
+        - src: workload-fio-script-cm.yml.j2
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/workload-fio-script-cm.yml"
+        - src: workload-env.yml.j2
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/workload-storage-fio-env.yml"
+
+    - name: Check if scale-ci-tooling namespace exists
+      shell: |
+        oc get projects | grep "scale-ci-tooling"
+      ignore_errors: true
+      changed_when: false
+      register: scale_ci_tooling_ns_exists
+    - name: print status of scale_ci_tooling_ns_exists
+      debug:
+        var: scale_ci_tooling_ns_exists
+
+    - name: Ensure any stale scale-ci-storage-fio job is deleted
+      shell: |
+        oc delete job scale-ci-storage-fio -n scale-ci-tooling
+      register: scale_ci_tooling_project
+      failed_when: scale_ci_tooling_project.rc == 0
+      until: scale_ci_tooling_project.rc == 1
+      retries: 60
+      delay: 1
+      when: scale_ci_tooling_ns_exists.rc == 0
+
+    - name: Block for non-existing tooling namespace
+      block:
+        - name: Create tooling namespace
+          shell: |
+            oc create -f {{ansible_user_dir}}/scale-ci-tooling/scale-ci-tooling-ns.yml
+      when: scale_ci_tooling_ns_exists.rc != 0
+
+    - name: Create tooling service account
+      shell: |
+        oc create serviceaccount useroot -n scale-ci-tooling
+        oc adm policy add-scc-to-user privileged -z useroot -n scale-ci-tooling
+      when: enable_pbench_agents|bool
+
+    - name: Add serviceaccount for useroot when pbench is not used
+      shell: |
+        oc create serviceaccount useroot -n scale-ci-tooling
+      when: not enable_pbench_agents|bool
+
+    - name: Create/replace kubeconfig secret
+      shell: |
+        oc replace --force -n scale-ci-tooling -f "{{ansible_user_dir}}/scale-ci-tooling/kubeconfig-secret.yml"
+
+    - name: Create/replace the pbench configmap
+      shell: |
+        oc replace --force -n scale-ci-tooling -f "{{ansible_user_dir}}/scale-ci-tooling/pbench-cm.yml"
+
+    - name: Create/replace pbench ssh secret
+      shell: |
+        oc replace --force -n scale-ci-tooling -f "{{ansible_user_dir}}/scale-ci-tooling/pbench-ssh-secret.yml"
+
+    - name: Create/replace workload script configmap
+      shell: |
+        oc replace --force -n scale-ci-tooling -f "{{ansible_user_dir}}/scale-ci-tooling/workload-fio-script-cm.yml"
+
+    - name: Create/replace workload script environment configmap
+      shell: | 
+        oc replace --force -n scale-ci-tooling -f "{{ansible_user_dir}}/scale-ci-tooling/workload-storage-fio-env.yml"
+
+    - name: Create/replace workload job to that runs workload script
+      shell: |
+        oc replace --force -n scale-ci-tooling -f "{{ansible_user_dir}}/scale-ci-tooling/workload-job.yml"
+
+    - name: Poll until job is active
+      shell: |
+        oc get job scale-ci-storage-fio -n scale-ci-tooling -o json
+      register: job_json
+      retries: 60
+      delay: 2
+      until: job_json.stdout | from_json | json_query('status.active==`1`')
+
+    - name: Poll until job is complete
+      shell: |
+        oc get job scale-ci-storage-fio -n scale-ci-tooling -o json
+      register: job_json
+      retries: "{{job_completion_poll_attempts}}"
+      delay: 60
+      until: job_json.stdout | from_json | json_query('status.succeeded==`1` || status.failed==`1`')
+      failed_when: job_json.stdout | from_json | json_query('status.succeeded==`1`') == false
+      when: job_completion_poll_attempts|int > 0
--- a/workloads/templates/workload-fio-script-cm.yml.j2
+++ b/workloads/templates/workload-fio-script-cm.yml.j2
@@ -1,0 +1,244 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: scale-ci-workload-script
+data:
+  run.sh: |
+    #!/bin/sh
+    set -eo pipefail
+    # pbench Configuration
+    echo "$(date -u) Configuring pbench for FIO I/O scale test"
+    mkdir -p /var/lib/pbench-agent/tools-default/
+    echo "${USER_NAME:-default}:x:$(id -u):0:${USER_NAME:-default} user:${HOME}:/sbin/nologin" >> /etc/passwd
+    echo "" > /var/lib/pbench-agent/tools-default/oc
+    echo "workload" > /var/lib/pbench-agent/tools-default/label
+    if [[ -v ENABLE_PBENCH_AGENTS ]]; then
+      echo "" > /var/lib/pbench-agent/tools-default/disk
+      echo "" > /var/lib/pbench-agent/tools-default/iostat
+      echo "" > /var/lib/pbench-agent/tools-default/mpstat
+      echo "" > /var/lib/pbench-agent/tools-default/perf
+      echo "" > /var/lib/pbench-agent/tools-default/pidstat
+      master_nodes=`oc get nodes -l pbench_agent=true,node-role.kubernetes.io/master= --no-headers | awk '{print $1}'`
+      for node in $master_nodes; do
+        echo "master" > /var/lib/pbench-agent/tools-default/remote@$node
+      done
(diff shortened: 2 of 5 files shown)


ekuric:
@akrzos @jmencak @chaitanyaenr  these [env vars](https://gist.github.com/ekuric/27c5ec52fa0bfe5d942c5dd3b2a9252d) I used for testing - I will clean up this more , but now will welcome one more functional test in case I missed something. Pbench-fio [issue](https://github.com/distributed-system-analysis/pbench/issues/1258)

workloads/templates/workload-fio-script-cm.yml.j2
```diff
@@ -0,0 +1,244 @@
...
+    echo "$(date -u) Done configuring pbench for FIO I/O scale test"
+    # End pbench Configuration
+    #
+    # Test Configuration
+    echo "$(date -u) Creating pods for FIO I/O scale test - pods will be in namespace:" {{ fiotest_basename }}0
+    pbench-user-benchmark --config="{{ fiotest_prefix }}-pods-{{ fiotest_maxpods }}-sc-{{ fiotest_storageclass }}-create_pods" -- 'VIPERCONFIG=/root/workload/fiotest.yml openshift-tests run-test "[Feature:Performance][Serial][Slow] Load cluster should load the cluster [Suite:openshift]"'
+    echo "$(date -u) Pods for FIO I/O test created."
+
+    # wait until all pods are started and then collect data
+    while [[ $(oc get pods -n {{ fiotest_basename }}0 | grep fiotest-pod | grep Run |wc -l) -lt {{ fiotest_maxpods }}  ]] ; do
```
akrzos, line 44: Maybe change `grep Run | wc -l` to `grep Run -c` ?

workloads/templates/workload-fio-script-cm.yml.j2
```diff
@@ -0,0 +1,244 @@
...
+      sleep 30
```
akrzos, line 45: Should we really sleep 30s each time.  Also I think we can wait for replicas to be ready using this same logic here:

https://github.com/openshift-scale/workloads/blob/master/workloads/files/workload-network-script-cm.yml#L70-L75
https://github.com/openshift-scale/workloads/blob/master/workloads/files/workload-network-script-cm.yml#L136-L153

workloads/templates/workload-fio-script-cm.yml.j2
```diff
@@ -0,0 +1,244 @@
...
+      creationTimestamp:
+      annotations:
+        description: Template to create ssh secret
+        tags: ''
+    objects:
+      - apiVersion: v1
+        kind: Secret
+        metadata:
+          name: pbench-ssh
+          namespace: {{ fiotest_basename }}0
```
akrzos, line 217: I am not sure the namespace is needed here as cluster loader should place this in the correct project/namespace.

workloads/templates/workload-fio-script-cm.yml.j2
```diff
@@ -0,0 +1,244 @@
...
+        template:
+          metadata:
+            labels:
+              name: fio
+            name: fio
+          spec:
+            hostNetwork: false
+            containers:
+              - name: "fio"
+                image: {{ fiotest_pod_image }}
```
akrzos, line 156: I think this should be ${STORAGE_FIO_IMAGE} and "templated" by the oc process command that cluster loader runs.

workloads/templates/workload-fio-script-cm.yml.j2
```diff
@@ -0,0 +1,244 @@
...
+        tags: fio,perf
+    labels:
+      name: fiotest-test
+    objects:
+    - apiVersion: v1
+      kind: PersistentVolumeClaim
+      metadata:
+        name: "${PVC_NAME}"
+        annotations:
+          volume.beta.kubernetes.io/storage-class: {{ fiotest_storageclass }}
```
akrzos, line 130: We should strive to use the oc process command "templating" when possible. (Ideally this reduces cognitive load with templating since there are multiple things templating here)

akrzos (changes requested):
Some initial feedback.  I will try it out on my cluster as well in the near future.

akrzos:
@ekuric looks like we ended up with a merge conflict as where here on the workload-job.yml.j2 file.  Please review so we can get this merged in.  Thanks!

workloads/templates/workload-fio-script-cm.yml.j2
portante: Perhaps even `grep -c Run`?

workloads/templates/workload-fio-script-cm.yml.j2
```diff
@@ -0,0 +1,244 @@
...
+    fi
+  fiotest.sh: |
+    #!/bin/bash
+    cp /root/.kube/config /tmp/config
+    export KUBECONFIG=/tmp/config
+    # Not sure is this necessary - re-test
+    sed -i 's/\-o StrictHostKeyChecking=no/\-o StrictHostKeyChecking=no \-l default \-p 2022/g' /opt/pbench-agent/base
+    # to do - fix pbench-fio to support custom options
+    # open issue why getconf.py returns empty string in container
+    curl -o /opt/pbench-agent/bench-scripts/pbench-fio https://raw.githubusercontent.com/ekuric/openshift/master/iotest/pbench-fio
```
portante, line 80: The container has `pbench-fio` installed, but over-writing the `pbench-fio` script seems a bit over kill just to get the `pbench-fio` version string.  That value is supposed to come from the `pbench-agent.cfg` file.

Every pbench-agent install has an `/opt/pbench-agent/config/pbench-agent.cfg` file.  So if this container image does not have that configuration file installed with a valid `pbench-fio` version string, then something is wrong with how the containers are built.

If it *does* have a `pbench-agent.cfg` file, then there is something else wrong with the installation of pbench-agent in the container.  Where is that install performed?  I assume when the container is built ...

portante: Seems like you could add a test at line 34 about to ensure the existence of the `pbench-agent.cfg` file.

workloads/templates/workload-fio-script-cm.yml.j2
```diff
@@ -0,0 +1,244 @@
...
+        echo "worker" > /var/lib/pbench-agent/tools-default/remote@$node
+      done
+    fi
+    source /opt/pbench-agent/profile
+    echo "$(date -u) Done configuring pbench for FIO I/O scale test"
+    # End pbench Configuration
+    #
+    # Test Configuration
+    echo "$(date -u) Creating pods for FIO I/O scale test - pods will be in namespace:" {{ fiotest_basename }}0
+    pbench-user-benchmark --config="{{ fiotest_prefix }}-pods-{{ fiotest_maxpods }}-sc-{{ fiotest_storageclass }}-create_pods" -- 'VIPERCONFIG=/root/workload/fiotest.yml openshift-tests run-test "[Feature:Performance][Serial][Slow] Load cluster should load the cluster [Suite:openshift]"'
```
portante, line 40: What does it mean to invoke `pbench-user-benchmark` here when this is a `pbench-fio` test?  This seems a bit odd: are we using `pbench-user-benchmark` to measure what it takes to *start* the "pods for FIO I/O scale" testing?  Are these pods used by the `pbench-fio` command in `fiotest.sh` below?

workloads/templates/workload-fio-script-cm.yml.j2
```diff
@@ -0,0 +1,244 @@
...
+       echo "Project {{ fiotest_basename }}0 is deleted ... test finished"
+    elif [[ {{ fiotest_cleanup }} == "false" ]]; then
+       echo "Test is done, but project {{ fiotest_basename }}0 is not be deleted due to FIOTEST_CLAENUP=false"
+    fi
+  fiotest.sh: |
+    #!/bin/bash
+    cp /root/.kube/config /tmp/config
+    export KUBECONFIG=/tmp/config
+    # Not sure is this necessary - re-test
+    sed -i 's/\-o StrictHostKeyChecking=no/\-o StrictHostKeyChecking=no \-l default \-p 2022/g' /opt/pbench-agent/base
```
portante, line 77: This line should be kept in the `pbench-agent.cfg` file on the image.  Why do we have to use a `sed` operation here to change pbench-agent code in the live image?

workloads/templates/workload-fio-script-cm.yml.j2
```diff
@@ -0,0 +1,244 @@
...
+    # to do - fix pbench-fio to support custom options
+    # open issue why getconf.py returns empty string in container
+    curl -o /opt/pbench-agent/bench-scripts/pbench-fio https://raw.githubusercontent.com/ekuric/openshift/master/iotest/pbench-fio
+    oc get pods -n {{ fiotest_basename }}0 -o wide | awk '{print $6}' |grep -v IP > /tmp/clients.txt
+    pbench-fio --test-types={{ fiotest_testtype }} --config="{{ fiotest_testtype }}-pods{{ fiotest_maxpods }}-storageclass-{{ fiotest_storageclass }}" \
+    --clients=$(cat /tmp/clients.txt | awk -vORS=, '{ print $1 }' | sed 's/,$/\n/') -b {{ fiotest_bs }} \
```
portante, line 83: Since at least 0.60 of pbench-agent, `pbench-fio` takes a `--client-file` option so that you can just give it the `/tmp/clients.txt` if that just has IPs (which it seems it does from a cursory inspection).

portante (changes requested):
Seems like we need to resolve the location of the `pbench-agent.cfg` file in the containers.

workloads/templates/workload-fio-script-cm.yml.j2
ekuric: ack, will replace this, thought I am not sure what ` grep -c ` does in background , but less characters for sure ;)

workloads/templates/workload-fio-script-cm.yml.j2
ekuric: no, as soon as all pods are ready  it will move on, wait can be anything - depends only how often we want to run `oc`

workloads/templates/workload-fio-script-cm.yml.j2
ekuric: agree, this can be removed

workloads/templates/workload-fio-script-cm.yml.j2
ekuric: pbench-fio was installed inside container with `yum pbench-fio -y ` during image build , so I expect it set all fine as for any case where pbench-fio is installed , will check for `/opt/pbench-agent/config/pbench-agent.cfg`

workloads/templates/workload-fio-script-cm.yml.j2
ekuric: this test needs pods to be created in advance - pods can be created either with pbench-user-benchmark or on some different way, I decided to collect also information about system during pod create phase. Once pods are in place,  `pbench-fio` is called.

workloads/templates/workload-fio-script-cm.yml.j2
ekuric: cool , will try will it consume it properly if there is list of IPs in file.

workloads/vars/fio.yml
```diff
@@ -0,0 +1,63 @@
...
+fiotest_prefix: "{{ lookup('env', 'FIOTEST_PREFIX')|default('fiotest', true) }}"
+fiotest_cleanup: "{{ lookup('env', 'FIOTEST_CLEANUP')|default(true, true)|bool|lower }}"
+fiotest_basename: "{{ lookup('env', 'FIOTEST_BASENAME')|default('fiotest', true) }}"
+fiotest_maxpods: "{{ lookup('env', 'FIOTEST_MAXPODS')|default(6, true)|int }}"
+fiotest_pod_image: "{{ lookup('env', 'FIOTEST_POD_IMAGE')|default('quay.io/openshift-scale/scale-ci-fio', true) }}"
+fiotest_stepsize: "{{ lookup('env', 'FIOTEST_STEPSIZE')|default(1, true)|int }}"
+fiotest_pause: "{{ lookup('env', 'FIOTEST_PAUSE')|default(0, true)|int }}"
+fiotest_storage_size: "{{ lookup('env', 'FIOTEST_STORAGE_SIZE')|default('1Gi',true) }}"
+fiotest_storageclass: "{{ lookup('env', 'FIOTEST_STORAGECLASS')|default('csi-rbd', true) }}"
+access_modes: "{{ lookup('env', 'ACCESS_MODES')|default('ReadWriteOnce',true) }}"
```
jmencak, line 46: Would it make sense to prefix `access_modes`/`ACCESS_MODES` by `fiotest_`/`FIOTEST_`?  I know this is not done for pvcscale test, but shouldn't it also be?

nit: `s/('ReadWriteOnce',true)/('ReadWriteOnce', true)/`

workloads/vars/fio.yml
```diff
@@ -0,0 +1,63 @@
...
+fiotest_bs: "{{ lookup('env', 'FIOTEST_BS')|default(4,true)|int }}"
+fiotest_filename: "{{ lookup('env', 'FIOTEST_FILENAME')|default('/mnt/pvcmount/f2', true) }}"
+fiotest_filesize: "{{ lookup('env', 'FIOTEST_FILESIZE')|default('1G', true) }}"
+fiotest_runtime: "{{ lookup('env', 'FIOTEST_RUNTIME')|default(60,true)|int }}"
+fiotest_direct: "{{ lookup('env', 'FIOTEST_DIRECT')|default(1, true)|int }}" 
+fiotest_iodepth: "{{ lookup('env', 'FIOTEST_IODEPTH')|default(32,true)|int }}" 
+fiotest_testtype: "{{ lookup('env', 'FIOTEST_TESTTYPE')|default('write',true) }}" 
+fiotest_samples: "{{ lookup('env', 'FIOTEST_SAMPLES')|default(1,true)|int }}"
+storage_fio_test_image: "{{ lookup('env', 'STORAGE_FIO_TEST_IMAGE')|default('quay.io/openshift-scale/scale-ci-fio', true) }}"
```
jmencak, line 55: It would be nice to have these and env vars consistent.  Again, change to all the variables from `storage_fio_test` to `fiotest_storage`?  Also, the environment variables too (`s/STORAGE_FIO_TEST_/FIOTEST_STORAGE_/`).

workloads/vars/fio.yml
```diff
@@ -0,0 +1,63 @@
...
+
+# FIO IO test workload specific parameters:
+fiotest_prefix: "{{ lookup('env', 'FIOTEST_PREFIX')|default('fiotest', true) }}"
+fiotest_cleanup: "{{ lookup('env', 'FIOTEST_CLEANUP')|default(true, true)|bool|lower }}"
+fiotest_basename: "{{ lookup('env', 'FIOTEST_BASENAME')|default('fiotest', true) }}"
+fiotest_maxpods: "{{ lookup('env', 'FIOTEST_MAXPODS')|default(6, true)|int }}"
+fiotest_pod_image: "{{ lookup('env', 'FIOTEST_POD_IMAGE')|default('quay.io/openshift-scale/scale-ci-fio', true) }}"
+fiotest_stepsize: "{{ lookup('env', 'FIOTEST_STEPSIZE')|default(1, true)|int }}"
+fiotest_pause: "{{ lookup('env', 'FIOTEST_PAUSE')|default(0, true)|int }}"
+fiotest_storage_size: "{{ lookup('env', 'FIOTEST_STORAGE_SIZE')|default('1Gi',true) }}"
```
jmencak, line 44: nit: `s/default('1Gi',true)/default('1Gi', true)/`

workloads/templates/workload-fio-script-cm.yml.j2
```diff
@@ -0,0 +1,244 @@
...
+    objects:
+    - apiVersion: v1
+      kind: PersistentVolumeClaim
+      metadata:
+        name: "${PVC_NAME}"
+        annotations:
+          volume.beta.kubernetes.io/storage-class: {{ fiotest_storageclass }}
+      spec:
+        accessModes:
+         - {{ access_modes }}
```
jmencak, line 133: As mentioned below, I believe `{{ fiotest_access_modes }}` would be more consistent.

workloads/fio.yml
```diff
@@ -0,0 +1,141 @@
...
+
+    - name: Copy workload files
+      copy:
+        src: "{{item.src}}"
+        dest: "{{item.dest}}"
+      with_items:
+        - src: scale-ci-tooling-ns.yml
+          dest: "{{ansible_user_dir}}/scale-ci-tooling/scale-ci-tooling-ns.yml"
+          # - src: workload-fio-script-cm.yml
+          # dest: "{{ansible_user_dir}}/scale-ci-tooling/workload-fio-script-cm.yml"
```
jmencak, line 28: Why do we have these two comments?

jmencak (commented):
Adding a few other comments others haven't picked up yet.

workloads/templates/workload-fio-script-cm.yml.j2
ekuric: This is fixed if `/workloads/templates/pbench-cm.yml.j2` is updated and added 
```
 [pbench-fio]
 version = 3.12
 histogram_interval_msec = 10000
```
and later in this script pointed to this new file `export CONFIG=/opt/pbench-agent/config/pbench-agent.cfg`

PR #60, opened 2019-07-12
All workloads use the workload environment configmap

akrzos:
This PR enables all workloads to take advantage of using the ansible templated workload-env configmap.


chaitanyaenr (approved):
LGTM

Issue #68, opened 2019-07-23
Add conformance test to the workloads framework


chaitanyaenr:
This will enable us to run conformance ( collect pbench if needed ) to check if the OpenShift cluster is sane.

PR #69, opened 2019-07-23
Add conformance to the workloads framework

chaitanyaenr:
This will enable us to run the e2e tests to validate and check if the
OpenShift cluster is sane.

### DIFF
--- a/workloads/files/workload-conformance-script-cm.yml
+++ b/workloads/files/workload-conformance-script-cm.yml
@@ -1,0 +1,63 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: scale-ci-workload-script
+data:
+  run.sh: |
+    #!/bin/sh
+    set -eo pipefail
+    workload_log() { echo "$(date -u) $@" >&2; }
+    export -f workload_log
+    workload_log "Configuring pbench for Conformance"
+    mkdir -p /var/lib/pbench-agent/tools-default/
+    echo "${USER_NAME:-default}:x:$(id -u):0:${USER_NAME:-default} user:${HOME}:/sbin/nologin" >> /etc/passwd
+    if [ "${ENABLE_PBENCH_AGENTS}" = true ]; then
+      echo "" > /var/lib/pbench-agent/tools-default/disk
+      echo "" > /var/lib/pbench-agent/tools-default/iostat
+      echo "workload" > /var/lib/pbench-agent/tools-default/label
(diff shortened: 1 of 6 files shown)


chaitanyaenr:
This requires kubectl to be baked into the scale-ci-workload image: https://github.com/openshift-scale/images/pull/17.

chaitanyaenr:
Fixes https://github.com/openshift-scale/workloads/issues/68.

workloads/files/workload-conformance-script-cm.yml
```diff
@@ -0,0 +1,63 @@
...
+  workload.sh: |
+    #!/bin/sh
+    set -o pipefail
+
+    result_dir=/tmp
+    if [ "${PBENCH_INSTRUMENTATION}" = "true" ]; then
+      result_dir=${benchmark_results_dir}
+    fi
+    start_time=$(date +%s)
+    export KUBECONFIG=/root/.kube/config; /usr/bin/openshift-tests run openshift/conformance/parallel || exit 0
```
akrzos, line 60: I am not sure I agree with this exit 0 statement here. I would rather we fail the workload job if a failure occurred but I also suspect we will encounter failures with the conformance tests at a very high rate thus we might not really want that.

Perhaps we can capture the exit code and determine later on failure or not instead of losing the exit code here or we also capture some machine process-able count of failures and maybe we exit 1 if that count is too high.

akrzos (commented):
Overall looks good.  I will test it tomorrow in the am, I do want to investigate the exit behavior of conformance and if we can count and maybe pass/fail based on number of conformance failures or percentage of failures.

chaitanyaenr:
@akrzos  pass/fail for conformance is tricky as there might be some critical failures even if the number of test failures are less or there might be more failures which are not actually blockers.

akrzos:
I do agree that while measuring and/or failing based on test failure count won't be a perfect indicator but it could throw up a red flag if something is not well.  Anyhow we can discuss that logic in further detail outside this pr.  I ran conformance and it took 79m on my cluster.  Thanks @chaitanyaenr for putting this one together

akrzos:
@chaitanyaenr can you add an entry to this [file](https://github.com/openshift-scale/workloads/blob/master/docs/README.md) with a link to the docs page for the test and then we are good to merge.

chaitanyaenr:
@akrzos added a link to the docs in the README.md. PTAL.

akrzos (approved):

PR #71, opened 2019-07-31
Use the concurrent golang cluster loader implementation

akrzos:


akrzos:
DNM - Do not merge yet.

We need some object waits at the end to correctly determine when the workload "completes"

akrzos:
FYI @chaitanyaenr

akrzos:
Once this lands (https://github.com/openshift/origin/pull/23550) we should have mastervertical using concurrent cluster loader in which the objects are waited for.

jmencak (approved):
Untested, but in line with recent cluster-loader changes.

PR #86, opened 2019-09-13
pbench_agent pods will now land on different AZs

mohit-sheth:
earlier by default both the pbench_agent pods used to land on AZ 'a'. Now they will land on AZ 'a' and AZ 'b' respectively. This will ensure the tests are always cross AZ 

@chaitanyaenr  @Deepthidharwar Please review

### DIFF
--- a/workloads/tooling.yml
+++ b/workloads/tooling.yml
@@ -108,7 +108,8 @@
 
     - name: Label two worker nodes
       shell: |
-        oc get nodes -l node-role.kubernetes.io/worker= --no-headers | head -n 2 | awk '{print $1}' | xargs -I % oc label nodes % --overwrite pbench_agent=true
+        oc get nodes -l failure-domain.beta.kubernetes.io/zone=us-west-2a | grep worker | head -n 1 | awk '{print $1}' | xargs -I % oc label nodes % --overwrite pbench_agent=true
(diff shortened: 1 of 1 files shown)


workloads/tooling.yml
```diff
@@ -108,7 +108,8 @@
 
     - name: Label two worker nodes
       shell: |
-        oc get nodes -l node-role.kubernetes.io/worker= --no-headers | head -n 2 | awk '{print $1}' | xargs -I % oc label nodes % --overwrite pbench_agent=true
+        oc get nodes -l failure-domain.beta.kubernetes.io/zone=us-west-2a | grep worker | head -n 1 | awk '{print $1}' | xargs -I % oc label nodes % --overwrite pbench_agent=true
```
chaitanyaenr, line 111: Finding the nodes based on region needs to be generic to all cloud platforms/regions.

chaitanyaenr (changes requested):

workloads/tooling.yml
Deepthidharwar: Please do not hard code the region. It needs to work in any cloud environment. For all you know there might me no VMs running in this particular region.
One way is to make sure is that region of the node picked first is different from that of the second node.

PR #88, opened 2019-10-24
Logging test automation added

RH-ematysek:


workloads/logging.yml
```diff
@@ -0,0 +1,82 @@
...
+        dest: workloads/
+    - name: Launch cluster-loader
+      shell: |
+        sudo podman run \
+        -v {{ ansible_facts['env']['KUBECONFIG']|default(ansible_facts['env']['HOME'] + '/.kube/config') }}:/root/.kube/config \
+        -v {{ansible_facts['env']['HOME']}}/workloads/:/root/workloads/ \
+        -i quay.io/openshift/origin-tests:{{ORIGIN_TESTS_VERSION}} \
+        /bin/bash -c 'export KUBECONFIG=/root/.kube/config && \
+        export VIPERCONFIG=/root/workloads/logtest.yml && \
+        openshift-tests run-test "[Feature:Performance][Serial][Slow] Load cluster should load the cluster [Suite:openshift]"'
```
sjug, line 69: You can use the concurrent cluster loader command `"[Feature:Performance][Serial][Slow] Load cluster concurrently with templates [Suite:openshift]"` if you're just deploying templates in your config (which seems to be the case).

RH-ematysek: When I change to that cluster-loader command and change my template (logtest-rc.json) from "apiVersion": "v1" to "apiVersion": "template.openshift.io/v1", cluster-loader no longer uses the parameters passed in logtest.yml, but instead uses the default values in logtest-rc.json. I'll have to look into this further.

RH-ematysek: See changes I am testing here:
https://github.com/RH-ematysek/workloads/commit/bf7a9a35b073cf1629354fbcde33c86dc39d5f0a

PR #99, opened 2020-01-08
Cast the value types explicitly in the job template

### LINKED ISSUE #97: Running workloads against non azure failing
Looks like https://github.com/openshift-scale/workloads/pull/96 introduced some bug, as the workload job seems to be failing as it can't find the azure_auth file, the only way to get around this right now is to edit the vars file for the respective workload and set azure_auth to false and the azure_auth_file to "". The job has the following log in description:

```
  Warning  FailedMount  1s (x8 over 65s)  kubelet, ip-10-0-176-169.us-west-2.compute.internal  MountVolume.SetUp failed for volume "azure-auth" : secret "azure-auth" not found
```

chaitanyaenr:
closes #97

### DIFF
--- a/workloads/templates/workload-job.yml.j2
+++ b/workloads/templates/workload-job.yml.j2
@@ -25,7 +25,7 @@ spec:
 {% if workload_job == "pvcscale" %}
       serviceAccountName: useroot
 {% endif %}
-{% if workload_job_privileged %}
+{% if workload_job_privileged| bool %}
       serviceAccountName: useroot
       hostNetwork: true
 {% endif %}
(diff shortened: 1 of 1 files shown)


aakarshg (approved):
odd ansible/jinja2 interaction that this pr is even needed but as long as it works we're good.

jtaleric (approved):
LGTM - did these vars not used to be boolean?

chaitanyaenr:
@jtaleric They are booleans but azure_auth var was behaving a bit weird like @aakarshg mentioned in https://github.com/openshift-scale/workloads/issues/97, this will explicitly cast it to boolean to resolve the issue.

PR #100, opened 2020-01-08
Ignore azure auth for scale and prometheus workloads

chaitanyaenr:
The azure authentication is not needed for scale, prometheus and
baseline workloads as they don't use cluster loader. This is needed as we
don't define the azure_auth vars in the scale and prometheus workloads
vars file.

### DIFF
--- a/workloads/templates/workload-env.yml.j2
+++ b/workloads/templates/workload-env.yml.j2
@@ -4,7 +4,6 @@ metadata:
   name: scale-ci-workload-{{workload_job}}-env
 data:
   ENABLE_PBENCH_AGENTS: "{{enable_pbench_agents|bool|lower}}"
-  AZURE_AUTH: "{{azure_auth|bool|lower}}"
 {% if workload_job == "http" %}
 {% for v in http_env_vars %}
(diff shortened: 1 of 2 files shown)


chaitanyaenr:
Added baseline workload to the list as well.

jtaleric (approved):
LGTM - as @chaitanyaenr mentioned on Slack. We ignore this var since we are not running cluster loader.

aakarshg (commented):
I understand that the var is not required as we don't run cluster loader for a set of tests, however at the same time if having the var being set is not an issue, i'd say let it be. This will help us when we add new tests that require running cluster loader as it wont require explicitly setting an additional var and figuring out logic.

chaitanyaenr:
@aakarshg the workloads which don't use the cluster loader will fail as we don't use/set the azure_auth. We need to ignore looking for azure_auth vars and mounting the secret for these workloads to work. The other way to make this work is to add the azure auth vars to the workloads vars file even though they don't use them which doesn't really make sense.

aakarshg (approved):
Understood, this looks good then merging this.

PR #103, opened 2020-01-10
Ignore azure auth for http workloads

mohit-sheth:
The azure authentication is not needed for http workloads as they don't use cluster loader.
@chaitanyaenr


chaitanyaenr (approved):
LGTM

PR #104, opened 2020-01-10
Restrict workloads on worker nodes

chaitanyaenr:
This commit will make sure all the objects created by the workloads
land on worker nodes instead of spreading across the cluster which
is a problem when there are infra nodes meant to run just
prometheus, router and registry.


jtaleric (approved):

chaitanyaenr:
Thanks for the review @jtaleric.

PR #105, opened 2020-01-11
workload objects will land on worker nodes

mohit-sheth:
This commit will make sure all the objects created by http workload land on worker nodes instead of spreading across the cluster @jtaleric @chaitanyaenr


chaitanyaenr (commented):
LGTM

PR #107, opened 2020-01-15
Skip azure auth for test workload

chaitanyaenr:
Test workload just creates a pod which sleeps forever, it doesn't
require azure auth to be passed as it doesn't run clusterloader.


mohit-sheth (approved):
LGTM

chaitanyaenr:
Thanks for the review @mohit-sheth.

Issue #111, opened 2020-02-02
DaemonSet, StatefulSet and Deployment kinds moved to apps/v1 API in Kube 1.16


chaitanyaenr:
The templates including the ones used by cluster loader and pbench agents/controller needs to be move to apps/v1 as @wabouhamad found out when running tests on OCP 4.4.

chaitanyaenr:
https://github.com/openshift-scale/workloads/pull/112 fixes it.

PR #112, opened 2020-02-04
Update pbench-infra-ds.yml pbench-master-ds.yml pbench-worker-ds.yml for OCP 4.4

wabouhamad:
This is to support template format and apiVersion for OCP 4.4 and kubernetes v1.17


chaitanyaenr:
Fixes #111

wabouhamad:
@mffiedler @chaitanyaenr PTAL

chaitanyaenr (approved):
LGTM, tested it on OCP 4.3 and it works fine. It should also work fine on OCP 4.4. Thanks.

PR #115, opened 2020-02-18
Avoid using infra nodes on OSD when selecting workers.

mdwn:
On OSD, infra nodes are also labeled as worker nodes by a technical
limitation. This change will make it so that the scale testing will only
select worker nodes that are not also labeled as infra nodes. This
should be transparent for regular scale testing as the infra nodes are
not labeled this way, but should hopefully maintain parity with the
intended functionality on OSD.

Note: Working on testing.

### DIFF
--- a/workloads/templates/pbench-worker-ds.yml.j2
+++ b/workloads/templates/pbench-worker-ds.yml.j2
@@ -62,6 +62,16 @@ spec:
         hostPath:
           path: /var/lib/kubelet/pods
       serviceAccountName: useroot
-      nodeSelector:
-        node-role.kubernetes.io/worker: ""
-        pbench_agent: "true"
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: node-role.kubernetes.io/worker
+                operator: Exists
+              - key: node-role.kubernetes.io/infra
+                operator: DoesNotExist
+              - key: pbench_agent
+                operator: In
+                values:
+                - "true"
--- a/workloads/network.yml
+++ b/workloads/network.yml
@@ -46,14 +46,14 @@
 
     - name: Label worker nodes with pbench agent for network tests
       shell: |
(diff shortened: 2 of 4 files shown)


mdwn:
@chaitanyaenr Any thoughts? We could use this for our testing -- merging it would prevent us from needing a custom branch of workloads or needing to tweak our OSD clusters from their defaults.

workloads/templates/pbench-worker-ds.yml.j2
```diff
@@ -62,6 +62,12 @@ spec:
         hostPath:
           path: /var/lib/kubelet/pods
       serviceAccountName: useroot
-      nodeSelector:
-        node-role.kubernetes.io/worker: ""
-        pbench_agent: "true"
+      affinity:
```
chaitanyaenr, line 65: @meowfaceman Thanks for the PR, we also need the logic to isolate the pbench-agent pods to only couple of nodes of interest since collecting the data from all the workers leads to duplication. We do it labeling the worker nodes of interest ( usually 2 nodes ) with pbench_agent=true and then isolate the pbench-agent worker daemonset to just those labeled nodes. The current logic deploys it on all the workers, we need to preserve the previous logic to avoid collecting data from all the worker nodes at large scale.

mdwn: Huh. So this is not the behavior that I was seeing prior to this change. Before this patch, on a cluster with 2 masters, 3 infra (also labeled workers), and 4 workers, the tooling workload tried to deploy 7 pbench workers. So that's with the code as is in the repo.

mdwn: I'll restore the pbench agent worker label though.

mdwn:
I still need to test this new commit, but it'll have to wait until tomorrow.

mdwn:
@chaitanyaenr While testing this out, I realized that a lot of the existing oc label commands would inadvertently label infra nodes on OSD clusters. I added in some changes here which should hopefully be transparent to other openshift-scale tests, but I'm testing this out these new changes as we speak.

chaitanyaenr (approved):
LGTM, thanks for the PR.

PR #117, opened 2020-02-27
Add support to the test workload to skip running on workload node

chaitanyaenr:
The test workload pod which basically just runs sleep infinity to allow
us to run benchmarks which are not automated tries to run on the dedicated
workload node created as part of the post-install ( day 2 operation ) by
default. This commit exposes the the workload node selector, taints and
running the pod in privileged/unpriveleged mode option to give the user
more flexibility to be able to run the workload even in the absence of
dedicated workload node.


mohit-sheth (approved):
lgtm

PR #119, opened 2020-03-04
Add explicit tag for image to use master

aakarshg:
master tag of workload image is built periodically and it seems
to be more reliable, so using an explicit tag for the image.

Co-authored-by: Joe Talerico <jtaleric@redhat.com>


jtaleric (approved):
LGTM

Issue #121, opened 2020-03-24
RFE: Pbench has no knowledge of the platform, causing issues with tool data


aakarshg:
When nodes are scaled up or down, the agent pods are rescheduled however, after running any workload the tool data can't be captured as pbench has no knowledge of the nodes that its supposed to collect from. 

There's a workaround for it, where we launch a test pod and run pbench-clear-tools and then reregister the tools. However it's not documented, so even if the problem is not fixed through playbooks would be helpful to document it.

PR #122, opened 2020-03-24
Clear pbench remotes before configuring

chaitanyaenr:
This commit fixes the issue where there might be stale remotes
left from the previous benchmark runs which might no longer exist
because of operations like scale down. pbench-clear-tools is run to
clear them before configuring the remotes for each benchmark run.

### DIFF
--- a/workloads/files/workload-conformance-script-cm.yml
+++ b/workloads/files/workload-conformance-script-cm.yml
@@ -12,6 +12,11 @@ data:
     mkdir -p /var/lib/pbench-agent/tools-default/
     echo "${USER_NAME:-default}:x:$(id -u):0:${USER_NAME:-default} user:${HOME}:/sbin/nologin" >> /etc/passwd
     if [ "${ENABLE_PBENCH_AGENTS}" = true ]; then
+      source /opt/pbench-agent/profile
+      # clear tools/remotes to make sure there are no invalid remotes
+      set +eo pipefail
+      pbench-clear-tools
+      set -eo pipefail
       echo "" > /var/lib/pbench-agent/tools-default/disk
       echo "" > /var/lib/pbench-agent/tools-default/iostat
       echo "workload" > /var/lib/pbench-agent/tools-default/label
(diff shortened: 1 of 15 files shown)


chaitanyaenr:
Fixes #121

rsevilla87:
I ran into this issue from a different path. In my case I reinstalled the tools and they were scheduled on different nodes.

rsevilla87 (changes requested):
I've tried to execute the workaround manually. And I found out that the  `pbench-clear-tools ` command returns !=0 when it's not able to clean tools from a remote node, so this change will make the workload to hang in those cases  since the bash interpreter is configured with the exit option on the line 8.

```
[root@ovn ~]# echo "master" > /var/lib/pbench-agent/tools-default/remote@1.2.3.4
[root@ovn ~]# pbench-clear-tools
running . /opt/pbench-agent/profile; pbench-clear-tools --group default  on 1.2.3.4
ssh: connect to host 1.2.3.4 port 22: Connection timed out
Removing the remote remote@1.2.3.4
[root@ovn ~]# echo $?
1
```

A workaround might be to disable exit during this command execution with `set +e`  and re-enable it later, or we might also move forward the  `set -eo pipefail` statement.

chaitanyaenr:
Nice catch @rsevilla87, modified the command to exit 0 explicitly to avoid hitting the issue when the remote host no longer exists for it clear tools. PTAL.

rsevilla87:
@chaitanyaenr the proposed workaround will also break as the `pipefail` bash option is set.
According to the official bash documentation:

_If set, the return value of a pipeline is the value of the last (rightmost) command to exit with a non-zero status, or zero if all commands in the pipeline exit successfully_


Considering that we'll need to disable pipefail before the command `pbench-clear-tools` i.e.
```bash
$ set +eo pipefail
$ pbench-clear-tools
$ set -eo pipefail
```
Note that exit 0 is no longer required as we've also disabled bash exit option.

Hope that helps ;)

chaitanyaenr:
@rsevilla87 it definitely helps, will make the change. Thanks.

chaitanyaenr:
@rsevilla87 Made the suggested changes. PTAL.

rsevilla87:
lgtm @chaitanyaenr

rsevilla87 (approved):

aakarshg (approved):
LGTM

PR #125, opened 2020-04-01
Adding azure auth to environment vars for deployments per ns

dry923:
Adding deployments per namespace section.


chaitanyaenr (approved):
LGTM

PR #127, opened 2020-04-16
client pod will land on workload node if it is present

mohit-sheth:
It is not possible to specify workload node during pipeline runs as there is a fresh install every time.
This commit will make sure that client pod/load_generator will land on workload node.

### DIFF
--- a/workloads/files/workload-http-script-cm.yml
+++ b/workloads/files/workload-http-script-cm.yml
@@ -301,8 +301,15 @@ data:
       local oc_whoami=`oc whoami`
 
       test "$HTTP_TEST_LOAD_GENERATOR_NODES" || {
-        echo "Not (un)labelling nodes, load generator nodes unspecified."
-        return 0
+        echo "Load generator nodes unspecified."
+        if [ $HTTP_TEST_LOAD_GENERATORS -eq 1 ] ; then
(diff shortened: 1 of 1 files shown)


chaitanyaenr (approved):
LGTM

workloads/files/workload-http-script-cm.yml
```diff
@@ -301,8 +301,15 @@ data:
       local oc_whoami=`oc whoami`
 
       test "$HTTP_TEST_LOAD_GENERATOR_NODES" || {
-        echo "Not (un)labelling nodes, load generator nodes unspecified."
-        return 0
+        echo "Load generator nodes unspecified."
+        if [ $HTTP_TEST_LOAD_GENERATORS -eq 1 ] ; then
+          HTTP_TEST_LOAD_GENERATOR_NODES=$(oc get nodes --no-headers | grep "workload" |  awk '{print $1}')
```
chaitanyaenr, line 306: Maybe add a check to see if workload node exists?

mohit-sheth: if workload node doesn't exist it will run like it used to, use a worker node.

chaitanyaenr: Ack, thanks.

PR #128, opened 2020-04-21
Nodevert heavy

rsevilla87:
Include a "heavy" nodevertical workload. Further info about this workload is available at the included documentation.

250 pods tested successfully in m5.2xlarge nodes (8vCPU and 32GiB). Results below:

```
root@ip-172-31-76-65: ~ # oc describe nodes -l nodevertical=true | grep Non-terminated
Non-terminated Pods:                      (250 in total)
Non-terminated Pods:                      (250 in total)
Non-terminated Pods:                      (250 in total)

root@ip-172-31-76-65: ~ # oc adm top nodes -l nodevertical=true
NAME                                        CPU(cores)   CPU%   MEMORY(bytes)   MEMORY%   
ip-10-0-143-79.us-west-2.compute.internal   755m         10%    8495Mi          28%       
ip-10-0-155-81.us-west-2.compute.internal   732m         9%     8810Mi          29%       
ip-10-0-166-82.us-west-2.compute.internal   753m         10%    9116Mi          30%       
```

The interesting part is that thanks to this workload we're gonna be able to  stress other components like network since the application readiness probe performs a periodic database query.

Note that the application container is currently available at my own quay project (quay.io/rsevilla/perfapp:latest), we should move it to the openshift-scale org in the future.

### DIFF
--- a/docs/nodevertical.md
+++ b/docs/nodevertical.md
@@ -150,8 +150,31 @@ Period of time (in seconds) that cluster loader will wait for pods to come up to
 Default: `600`  
 Pass/fail criteria. Value to determine if NodeVertical workload executed in duration expected.
 
-## Smoke test variables
+### NODEVERTICAL_HEAVY
+Default: `false`
+Trigger nodevertical heavy workload. Rather than deploy sleep pods throughout all labeled nodes, this workload deploys a two layer application composed by a PostgreSQL database pod and another pod wih simple application that is able to perform several operations and store a result in the database. [Heavy node vertical app](https://github.com/rsevilla87/perfApp)
 
+### NODEVERTICAL_HEAVY_PROBE_ENDPOINT
+Default: `/ready`
+Readiness probe endpoint for the application deployed by the heavy nodevertical. The default `/ready` endpoint inserts a record with the current timestamp in the ts table of the DDBB. i.e. `INSERT INTO ts VALUES ('2006-01-02T15:04:05Z07:00')`
+
+Be aware that using nodevertical heavy has an extra resource consumption compared with the standard workload, 
+the following capture provides an idea about how much resources are required to run it.
+
+Results obtained from running 250 pods (client + database) with the default `/ready` endpoint.
+```
+root@ip-172-31-76-65: ~ # oc adm top nodes -l nodevertical=true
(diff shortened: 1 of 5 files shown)


chaitanyaenr (commented):
This is awesome @rsevilla87. Just one suggestion, maybe add a note in the docs on the resource ( cpu, memory ) requirements, this way the user can turn on the heavy workload when the workers nodes are big enough? Thoughts?

rsevilla87:
> This is awesome @rsevilla87. Just one suggestion, maybe add a note in the docs on the resource ( cpu, memory ) requirements, this way the user can turn on the heavy workload when the workers nodes are big enough? Thoughts?

I could add the above results to give the user an idea of the resource consumption. What do you think?
In the other hand, I'm also working in migrating this workloads to deployments as discussed in other thread.

chaitanyaenr:
@rsevilla87 sounds good.

rsevilla87:
@chaitanyaenr I've added some info about memory consumption at the docs as well as replaced DeploymentConfigs for Deployments as stated at https://github.com/openshift-scale/workloads/issues/130

docs/nodevertical.md
```diff
@@ -150,8 +150,27 @@ Period of time (in seconds) that cluster loader will wait for pods to come up to
...
+Default: `false`
+Trigger nodevertical heavy workload. Rather than deploy sleep pods throughout all labeled nodes, this workload deploys a two layer application composed by a PostgreSQL database pod and another pod wih simple application that is able to perform several operations and store a result in the database. [Heavy node vertical app](https://github.com/rsevilla87/perfApp)
+
+### NODEVERTICAL_HEAVY_PROBE_ENDPOINT
+Default: `/ready`
+Readiness probe endpoint for the application deployed by the heavy nodevertical. The default `/ready` endpoint inserts a record with the current timestamp in the ts table of the DDBB. i.e. `INSERT INTO ts VALUES ('2006-01-02T15:04:05Z07:00')`
 
+Results obtained of running 2 pods (client + database) with the default `/ready` endpoint.
+```
+# oc adm top pods -n nodevertical
```
chaitanyaenr: Maybe add a note for the users to use this info to check if they have enough resources on the worker nodes to host 250 similar pods before running the nodevert heavy? Thoughts?

chaitanyaenr (approved):
LGTM

PR #131, opened 2020-04-24
Add conprof collection for apiserver/prom to nodevert

aakarshg:
This is an initial draft to collect the pprof data for apiserver and prometheus through conprof when running nodervert. Once, this is merged will make same changes for other workloads

### DIFF
--- a/workloads/templates/workload-nodevertical-script-cm.yml.j2
+++ b/workloads/templates/workload-nodevertical-script-cm.yml.j2
@@ -44,7 +44,13 @@ data:
     envsubst < /root/workload/nodevertical.yaml.template > /tmp/nodevertical.yaml
     envsubst < /root/workload/nodevert-pod-labeled.yaml.template > /tmp/nodevert-pod-labeled.yaml
     workload_log "Done configuring NodeVertical test"
-
+    if [ "${PPROF_COLLECT}" = "true" ]; then
+      workload_log "Configuring conprof"
+      envsubst < /root/workload/conprof.yaml.template > /tmp/conprof.yaml
+      envsubst < /root/workload/conprof_start.sh > /tmp/conprof_start.sh
+      envsubst < /root/workload/conprof_stop.sh > /tmp/conprof_stop.sh
+      workload_log "Done configuring conprof"
+    fi
     workload_log "Running NodeVertical workload"
     if [ "${PBENCH_INSTRUMENTATION}" = "true" ]; then
       pbench-user-benchmark -- sh /root/workload/workload.sh
@@ -76,11 +82,23 @@ data:
     # TODO: Check pbench-agent collected metrics for Pass/Fail
     # TODO: Check prometheus collected metrics for Pass/Fail
     workload_log "Test Analysis: Passed"
+  conprof_start.sh: |
+    #!/bin/sh
+    set -o pipefail
+    nohup /usr/bin/conprof all --config.file /tmp/conprof.yaml --log.level=debug --storage.tsdb.path=/tmp/data &>/tmp/conprof.log &
+  conprof_stop.sh: |
+    #!/bin/sh
+    set -o pipefail
+    pkill conprof
   workload.sh: |
     #!/bin/sh
     set -o pipefail
 
     result_dir=/tmp
+    if [ "${PPROF_COLLECT}" = "true" ]; then
+      workload_log "Starting conprof"
+      bash /tmp/conprof_start.sh
+    fi
     if [ "${PBENCH_INSTRUMENTATION}" = "true" ]; then
       result_dir=${benchmark_results_dir}
     fi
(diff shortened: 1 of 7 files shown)


workloads/templates/workload-nodevertical-script-cm.yml.j2
```diff
@@ -72,11 +78,23 @@ data:
...
     # TODO: Check prometheus collected metrics for Pass/Fail
     workload_log "Test Analysis: Passed"
+  conprof_start.sh: |
+    #!/bin/sh
+    set -o pipefail
+    nohup /usr/bin/conprof all --config.file /tmp/conprof.yaml --log.level=debug --storage.tsdb.path=/tmp/data &>/tmp/conprof.log &
+  conprof_stop.sh: |
+    #!/bin/sh
+    set -o pipefail
+    ps -ef | grep "conprof all --config.file /tmp/conprof.yaml --log.level=debug" | grep -v grep | awk '{print $2}' | xargs kill
```
rsevilla87: A `pkill conprof` should be enough to kill all conprof instances.

workloads/templates/workload-nodevertical-script-cm.yml.j2
```diff
@@ -143,3 +168,121 @@ data:
...
+  conprof.yaml.template: |
+    scrape_configs:
+    - job_name: 'apiserver0'
+      scrape_interval: 30s
+      scrape_timeout: 10m
+      scheme: https
+      tls_config:
+        insecure_skip_verify: true
+      static_configs:
+      - targets: ['apiserver0-openshift-kube-apiserver.apps.{{clustername}}.perf-testing.devcluster.openshift.com']
```
rsevilla87: We might specify multiple targets here to avoid repeating ourselves, i.e
```
- targets: ['apiserver0-openshift-kube-apiserver.apps.{{clustername}}.perf-testing.devcluster.openshift.com', 'apiserver1-openshift-kube-apiserver.apps.{{clustername}}.perf-testing.devcluster.openshift.com', 'apiserver2-openshift-kube-apiserver.apps.{{clustername}}.perf-testing.devcluster.openshift.com'] 
```

In this case we should also make the the job_name more generic with something like apiserver .

rsevilla87 (commented):
Nice job @aakarshg :) I've added a couple of minor comments

workloads/templates/workload-nodevertical-script-cm.yml.j2
aakarshg: ooh nice, i'd try that and use that instead...

workloads/templates/workload-nodevertical-script-cm.yml.j2
aakarshg: its actually better to avoid specifying multiple targets, as we can then use specific labels for each in this case i'm trying to apply label node=node_name ( might not happen in this pr but maybe subsequent ones ) and then also add a label called apiserver=true. Having more detailed labels will help us easily get results, vs trying to figure out which profile is associated with which master.

jtaleric (commented):
Should we add a check for both the start and stop, to ensure the process has indeed started and stopped? It seems like we have blind faith in things running :)

workloads/vars/nodevertical.yml
```diff
@@ -32,6 +32,9 @@ pbench_ssh_private_key_file: "{{ lookup('env', 'PBENCH_SSH_PRIVATE_KEY_FILE')|de
 pbench_ssh_public_key_file: "{{ lookup('env', 'PBENCH_SSH_PUBLIC_KEY_FILE')|default('~/.ssh/id_rsa.pub', true) }}"
 pbench_server: "{{ lookup('env', 'PBENCH_SERVER')|default('', true) }}"
 
+# pporf variables
+pprof_collect: "{{ lookup('env', 'PPROF_COLLECT')|default(false, true)|bool|lower }}"
```
chaitanyaenr, line 36: The docs needs be updated with the details about the pprof collection.

workloads/templates/workload-nodevertical-script-cm.yml.j2
```diff
@@ -143,3 +168,121 @@ data:
...
+          trace:
+            enabled: false
+    - job_name: 'apiserver1'
+      scrape_interval: 30s
+      scrape_timeout: 10m
+      scheme: https
+      tls_config:
+        insecure_skip_verify: true
+      static_configs:
+      - targets: ['apiserver1-openshift-kube-apiserver.apps.{{clustername}}.perf-testing.devcluster.openshift.com']
```
chaitanyaenr: Do we have to modify the target when using a different base domain instead of perf-testing.devcluster.openshift.com or it doesn't matter?

workloads/templates/workload-nodevertical-script-cm.yml.j2
```diff
@@ -98,6 +116,14 @@ data:
     end_time=$(date +%s)
     duration=$((end_time-start_time))
 
+    if [ "${PPROF_COLLECT}" = "true" ]; then
+      workload_log "Stopping conprof"
+      bash /tmp/conprof_stop.sh
+      cp /tmp/conprof.log ${result_dir}/conprof.log
+      cp /tmp/conprof.yaml ${result_dir}/conprof.yaml
+      tar -czvf ${result_dir}/conprof.tar.gz /tmp/data/
```
chaitanyaenr, line 124: Maybe add a step in the playbook to run `$ oc cp ` to copy the data from /tmp of the pod in case pbench_instrumentation is not set? This way it works with both pbench enabled/disabled. Thoughts?

aakarshg: @chaitanyaenr that's a nice idea, do we currently do this with any other workload to copy data? if so, can you please point me to it, so that i can try to adapt it here. I've a few concerns about trying to do this, especially around how do we orchestrate job completion from outside the pod, but if there's a way we're doing this already then i can just try to do the same.

chaitanyaenr (approved):
LGTM

workloads/templates/workload-nodevertical-script-cm.yml.j2
chaitanyaenr: Resolving the conversation as per the discussion to implement a bottle based server for handling the data to work even when pbench is disabled. Thanks.

rsevilla87 (approved):
LGTM!

PR #133, opened 2020-04-29
source the pbench-agent/profile

mohit-sheth:
we ship the results to pbench server even when pbench-agents are not enabled.

### DIFF
--- a/docs/conformance.md
+++ b/docs/conformance.md
@@ -49,7 +49,7 @@ Location of kubeconfig on orchestration host.
 
 ### PBENCH_INSTRUMENTATION
 Default: `false`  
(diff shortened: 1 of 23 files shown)


aakarshg (approved):
not sure if this is a feature, but either way the implications of setting the vars to true vs false does need to be documented well on.

chaitanyaenr:
@aakarshg agreed, maybe we can make the docs ( https://github.com/openshift-scale/workloads/blob/master/docs/nodevertical.md#pbench_instrumentation ) more explicit by adding a note that pbench can be used without enabling agents and the results can be shipped to the server.

mohit-sheth:
@chaitanyaenr ^ Done

chaitanyaenr (approved):
LGTM

PR #135, opened 2020-05-11
Minor improvement to workload pod logs for nodevert

### LINKED ISSUE #134: Improve logging when clusterloader fails during nodevert tests
Irrespective of whether clusterloader failed to launch the required number of pods or the actual run duration exceeded expected duration, the exit code from the script is 1, for example at https://github.com/openshift-scale/workloads/blob/master/workloads/templates/workload-nodevertical-script-cm.yml.j2#L73

For new user this maybe misleading as even though the test executed successfully (required number of pod were created), the workload pod is marked as `Error` due to the return code. As a first step we can improve logging in the script to make it more explicit.

smalleni:
Fixes: #134


chaitanyaenr (approved):
LGTM. The user will be able to know the exact reason for failure with this change.

PR #140, opened 2020-06-03
Add a note about pbench deprecation

chaitanyaenr:
This commit adds a note about the pbench integration not being
actively maintained/supported in the Workloads repository. It can
still be used as is if it works without any issues. All the workloads
are being actively maintained/supported and it's recommended
to use Prometheus metrics to analyze the results.


jtaleric (approved):
LGTM

PR #143, opened 2020-07-09
Add support to promethues-scale to collect conprof data

mohit-sheth:
similar to pod-density and cluster-density PRs


chaitanyaenr (approved):
LGTM

PR #147, opened 2020-09-16
add explicit datatypes

mohit-sheth:
the older index had everything as a string


jtaleric (changes requested):
if we are only changing due to the mapping type, we can fix that in the index.... Do you have a mapping you want?

jtaleric (approved):
LGTM

